PR #67, opened 2026-03-02
fix: correct gateway auth format and model display logic

TGLTommy:
## Summary

- **`src/lib/websocket.ts`**: Fix gateway WebSocket handshake — OpenClaw gateway requires `auth: { token }` not `auth: { password }`, causing all connection attempts to fail with `INVALID_REQUEST`.

- **`src/components/panels/session-details-panel.tsx`**: Fix `getModelInfo()` always returning the first model catalog entry (`haiku`) for any unrecognized model name. The bug: `providerAliases` (a truthy string) was used directly as a `.find()` predicate, so every model in the array matched and the first entry was always returned. Fixed by comparing `m.alias === matchedAlias` instead.

## Root Cause (model display bug)

```js
// Before: providerAliases is a truthy string → matches every element → returns first (haiku)
availableModels.find(m =>
  m.name === modelName ||
  m.alias === modelName ||
  providerAliases          // ← always truthy if any alias substring matched
)

// After: precise comparison
availableModels.find(m =>
  m.name === modelName ||
  m.alias === modelName ||
  m.alias === matchedAlias  // ← correct
)
```

## Test plan

- [ ] Connect Mission Control to an OpenClaw gateway with token auth — gateway status should show **Online**
- [ ] Sessions using non-Anthropic models (e.g. `deepseek-ai/DeepSeek-V3.2`) should display the correct model name, not `haiku`

### DIFF
--- a/src/components/panels/session-details-panel.tsx
+++ b/src/components/panels/session-details-panel.tsx
@@ -32,13 +32,14 @@ export function SessionDetailsPanel() {
   const [expandedSession, setExpandedSession] = useState<string | null>(null)
 
   const getModelInfo = (modelName: string) => {
-    const modelAliases = availableModels.map(m => m.alias)
-    const providerAliases = modelAliases.find(alias => modelName.toLowerCase().includes(alias.toLowerCase()))
-    
-    return availableModels.find(m => 
-      m.name === modelName || 
+    const matchedAlias = availableModels
+      .map(m => m.alias)
+      .find(alias => modelName.toLowerCase().includes(alias.toLowerCase()))
+
+    return availableModels.find(m =>
+      m.name === modelName ||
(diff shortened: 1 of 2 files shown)


0xNyk:
Hey @TGLTommy — thanks for the contribution! Both fixes are valid catches:

1. **WebSocket auth field** (`password` → `token`) — this was already fixed in PR #54 (merged as commit 2eec86c), so main already has this fix.

2. **Model display bug** (`providerAliases` truthy predicate) — great catch! This was still present on main. I've cherry-picked your fix into PR #69 (now merged) with credit to you.

Since both fixes are now on main, I'm going to close this PR. Thanks again for identifying these bugs — the model display one especially was subtle and would have gone unnoticed otherwise. 🙏

PR #121, opened 2026-03-03
feat(#109): LLM selection dropdown for agents/cron jobs + LLM catalog panel

bhavikprit:
## Summary

Adds LLM discovery, a catalog panel for browsing available models, and a reusable dropdown selector for configuring models in agents and cron jobs.

### Changes

**API: `/api/llms`**
- Auto-discovers LLMs from 4 providers:
  - **Ollama** (local) — queries `/api/tags` endpoint
  - **LM Studio** (local) — queries OpenAI-compatible `/v1/models`
  - **OpenAI** (remote) — queries `/v1/models` if `OPENAI_API_KEY` set
  - **Anthropic** (remote) — known model list if `ANTHROPIC_API_KEY` set
- Returns provider status (online/offline), model metadata (size, parameters, quantization, family)
- Configurable via env vars: `OLLAMA_HOST`, `LM_STUDIO_HOST`

**New Panel: `LLMCatalogPanel`** (`src/components/panels/llm-catalog-panel.tsx`)
- Provider status cards with online/offline indicators
- Searchable model grid with provider badges
- Filter by provider (click provider card to filter)
- Click-to-view model detail modal with all metadata

**New Component: `LLMSelector`** (`src/components/ui/llm-selector.tsx`)
- Searchable dropdown for selecting LLM models
- Fetches available models from `/api/llms` on mount
- Provider badges on each option
- "None (default)" clear option
- Reusable across any form that needs model selection

**Integration**
- Added `LLMSelector` to cron job creation form in `CronManagementPanel`
- Added `LLMIcon` and `llm-catalog` nav item in OBSERVE group
- Registered `llm-catalog` route in `page.tsx`

## Risk Level
Low — new API route, new panel, new component. No changes to existing behavior.

## Tests
- `pnpm typecheck` — passes
- `pnpm lint` — passes
- `pnpm test` — 60/60 unit tests pass

## Contribution Checklist
- [x] Tests added/updated for behavior changes
- [x] Lint/typecheck/build passing
- [ ] Security review done if auth/data/crypto touched (N/A)
- [ ] DB migration tested if schema changed (N/A)

## Environment Variables
| Variable | Default | Description |
|---|---|---|
| `OLLAMA_HOST` | `http://127.0.0.1:11434` | Ollama API endpoint |
| `LM_STUDIO_HOST` | `http://127.0.0.1:1234` | LM Studio API endpoint |
| `OPENAI_API_KEY` | — | Enables OpenAI model discovery |
[...]

PR #112, opened 2026-03-03
feat: phase 1 workspace isolation across auth and core APIs

0xNyk:
## Summary\n- add migration v021 for workspace isolation primitives and indexes\n- add workspace context to auth sessions/users and expose workspace_id in auth responses\n- scope core task/agent/activity/notification/quality-review routes by workspace_id\n- propagate workspace_id through db helper activity/notification writes\n- update db helper unit tests for workspace-aware helper signatures\n\n## Validation\n- pnpm typecheck\n- pnpm lint\n- pnpm test

PR #126, opened 2026-03-03
feat(#109): LLM selection dropdown + LLM catalog panel

### LINKED ISSUE #109: LLM selection for each Agents/CronJobs
Currently is difficult to set the LLMs for each task in OpenClaw.
It would be great to have a Drop Down LLM selection in each place where a LLM can be used.

For example if I select an Agent or CronJon would be cool to have a drop down selection of the local models installed with Ollama or lllama.cpp, vllm... etc.

Also to have on the left bar, to have a LLM section where one can check out all LLMs

bhavikprit:
LLM discovery API, catalog panel, reusable selector dropdown, and integration into cron job form. Supersedes #121 (rebased onto latest main). Closes #109


0xNyk:
Not merging: overlaps with merged model selection work in PR #112 and changes behavior scope (new external provider polling endpoint/UI) without dedicated regression coverage.

Issue #217, opened 2026-03-05
[Bug] Critical Security Flaw: Agent Identity Can Be Spoofed


arana198:
## Describe the Bug
The current API design allows any authenticated user to impersonate any agent by simply modifying a request header. This creates a serious security and billing integrity issue.

Below is a breakdown of the problem and a proposed solution.

## Current Problem

```
┌─────────┐                    ┌──────────────────┐
│ OpenClaw│                    │ Mission Control  │
└────┬────┘                    └────────┬─────────┘
     │                                  │
     │ POST /api/tasks                  │
     │ Authorization: Bearer USER_KEY   │
     │ X-Agent-Name: "agent-123"        │
     │──────────────────────────────→   │
     │                                  │
     │                     ✅ User valid?
     │                     ⚠️ Trust header
     │                     (ANY user can claim ANY agent!)
     │                                  │
     │        ← 201 Task created        │
     │ ← attributed to "agent-123"      │
```

### Why This Fails

- User Alice authenticates using her API key.
- Alice sends a request with:
- X-Agent-Name: "admin-agent"
- The server trusts the header.
- The task is attributed to admin-agent, even though Alice created it.
- Alice can impersonate any agent in the system.

### Impact

This creates several serious problems:

- Agent impersonation
- Falsified work history
- Broken cost attribution
- Billing inaccuracies
- No cryptographic proof of agent identity

In short: agent identity is completely trust-based instead of cryptographically verified.

## Proposed Solution: Agent-Specific Credentials

When an agent registers, Mission Control should issue a signed credential. (This can be in Agents page)

Agents then authenticate using the issued token:

```
POST /api/tasks
Authorization: Bearer eyJhbGc...
Content-Type: application/json
```

```
{
  "title": "Summarize document",
  "description": "...",
  "priority": "high"
}
```

### Server Verification

Mission Control verifies:
- JWT signature is valid (signed by MC)
- Token has not expired
- Workspace matches
- Agent is still active

If valid:
→ Agent identity is trusted
→ Task is attributed correctly

0xNyk:
Status update: active fix is in PR #224 (server-side actor identity hardening). Current blocker is failing quality gate on that PR; once green + approved, this issue can be closed.

0xNyk:
Update: pushed a follow-up fix commit () to PR #224 to resolve the typecheck gate failure ( schema assertion drift). Quality gate is re-running now.

0xNyk:
[...]

Issue #216, opened 2026-03-05
[Bug] Agent Modal Messaging UI is Confusing — Unclear Purpose of “From” Field and Overall Layout


arana198:
After the implementation of this ticket:
https://github.com/builderz-labs/mission-control/issues/165

I expected the Agent modal UI to look and behave differently. In its current state, the interface feels somewhat cluttered and the interaction model is not very clear.

<img width="1026" height="917" alt="Image" src="https://github.com/user-attachments/assets/c60cdccd-52c3-453a-a057-eb8d94809ee1" />

## Issues Observed

- **Agent Modal UI Feels Unpolished** - The overall layout of the Agent modal does not appear very clean or intuitive. Some elements feel unnecessary or unclear in their purpose, which affects usability.
- **Confusing "From" Field in the Message Section** - The message section includes a "From" field. The purpose of this field is unclear.

If a user is sending a message to an agent from the Mission Control UI, the sender should automatically be the authenticated user. In that case, the system already knows who the sender is.

This raises a few questions:

- Why is the "From" field required?
- Can users impersonate other users or agents by modifying this field?
- Should the sender not simply default to the currently authenticated user?

Requiring users to manually specify the sender introduces unnecessary friction and may also create potential identity or impersonation concerns.

## Expected Behaviour

- Messages sent from the UI should automatically use the authenticated user as the sender.
- The UI should avoid fields that require manual identity selection when that information is already known.
- The Agent modal should have a simpler and cleaner messaging interface.

0xNyk:
Implemented in PR #229: removed manual sender selection in Agent Comms UI and now sender is automatically the authenticated user.\n\nValidation:\n- pnpm lint\n- pnpm typecheck\n- pnpm test:e2e (202 passed)

0xNyk:
Linked PRs: #229 directly fixes the Agent Comms sender UX ('From' field removal/simplification), and #224 adds server-side identity enforcement for sender attribution hardening.

PR #224, opened 2026-03-05
fix(security): enforce server-side actor identity

0xNyk:
## Summary
This is a security hardening step that removes client-controlled identity fields from core mutation flows.

### Changes
- tasks: ignore client `created_by`, always set actor server-side from authenticated user
- comments: ignore client `author`, always set author server-side from authenticated user
- direct messages: ignore client `from`, resolve sender server-side from authenticated user
- agent modal UI: remove editable "From" field to match server behavior
- validation: remove `created_by` / `author` / `from` from create schemas so callers cannot set them

### Security impact
- prevents simple impersonation via request body fields in task creation/comment/message flows
- reduces attribution spoofing and audit/billing integrity risk

## Tests
- `pnpm lint` ✅
- `pnpm test:e2e` ✅ (204 passed, 0 failed)
- added e2e regressions:
  - task create ignores spoofed `created_by`
  - task comment ignores spoofed `author`

## Issue linkage
- related to #217
- related to #216

### DIFF
--- a/src/app/api/agents/[id]/diagnostics/route.ts
+++ b/src/app/api/agents/[id]/diagnostics/route.ts
@@ -85,7 +85,7 @@ export async function GET(
     }
 
     const { searchParams } = new URL(request.url);
-    const requesterAgentName = (request.headers.get('x-agent-name') || '').trim();
+    const requesterAgentName = auth.user.agent_name?.trim() || '';
     const privileged = searchParams.get('privileged') === '1';
     const isSelfRequest = (requesterAgentName || auth.user.username) === agent.name;
 
--- a/src/app/api/agents/[id]/keys/route.ts
+++ b/src/app/api/agents/[id]/keys/route.ts
@@ -1,0 +1,234 @@
+import { createHash, randomBytes } from 'crypto'
+import { NextRequest, NextResponse } from 'next/server'
+import { requireRole } from '@/lib/auth'
+import { getDatabase } from '@/lib/db'
+import { logger } from '@/lib/logger'
+
+const ALLOWED_SCOPES = new Set([
+  'viewer',
+  'operator',
+  'admin',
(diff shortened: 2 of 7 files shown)


arana198:
I don’t think sending a user token together with an agent name is a good approach. It tightly couples user authentication with agent identification, which introduces unnecessary security and architectural concerns.

In the case of OpenClaw, this would require storing a user credential with privileged access and providing it to the agent runtime. That increases the risk surface and creates an undesirable dependency on user-level authentication for agent operations.

Would it make more sense for agents to authenticate using dedicated API keys instead?

- Ideally, agent API keys should:
- Be independent of user credentials
- Be revocable
- Support expiration

Be scoped specifically for agent-level permissions
This would provide a clearer separation between user identity and agent identity, while also improving overall security and operational control.

arana198:
You just need to extend the current AuthZ to be able to support agent as well as user authentication and authorisation

0xNyk:
Implemented in this same PR branch:

- Added dedicated agent API key auth path (separate from user/session auth)
- Added DB migration 027_agent_api_keys with scoped permissions, expiration, revocation, and usage timestamps
- Added admin lifecycle API for keys: GET/POST/DELETE /api/agents/{id}/keys
  - create returns one-time raw key (mca_...)
  - list returns metadata only
  - delete revokes by key_id
- Updated diagnostics self-scope identity source to use authenticated principal (auth.user.agent_name)
- Updated proxy auth gate to pass dedicated agent key candidates (mca_...) through to route-level DB validation

Validation run on this branch:
- pnpm lint
- pnpm typecheck
- pnpm build
- pnpm test:e2e -> 208 passed

This addresses the review direction to support agent authentication/authorization directly (not user token + agent header coupling), with revocable/expiring/scoped keys.

PR #283, opened 2026-03-11
fix(docker): allow check-node-version.mjs in .dockerignore for local builds

### LINKED ISSUE #282: Bug: Docker build fails with Cannot find module '/app/scripts/check-node-version.mjs'
**Describe the bug**
When building the Docker image locally from the source, the build fails at the `pnpm build` step. 
The `package.json` build script triggers `pnpm run verify:node`, which executes `node scripts/check-node-version.mjs`. However, because the `scripts` directory is excluded in `.dockerignore`, the file is not copied into the Docker context, leading to a `MODULE_NOT_FOUND` error.

**To Reproduce**
1. Clone the repository.
2. Run `docker compose build` or `docker build .`
3. See error:
   ```text
   > mission-control@2.0.0 verify:node /app 
   > node scripts/check-node-version.mjs 
   
   node:internal/modules/cjs/loader:1386 
   Error: Cannot find module '/app/scripts/check-node-version.mjs' 
   ```

**Expected behavior**
The Docker image should build successfully. The required build scripts should be available inside the container during the build phase.

**Proposed Solution**
Add an exception in `.dockerignore` to include the required script while keeping the rest of the `scripts/` directory ignored:
```dockerignore
!scripts/check-node-version.mjs
```

rainive:
## Description
Fixes #282

Building the Docker image locally currently fails because the `scripts` directory is ignored in `.dockerignore`, but the `build` command in `package.json` requires `scripts/check-node-version.mjs` to run the node version verification.

This PR adds an exclusion (`!scripts/check-node-version.mjs`) to `.dockerignore` so that only the necessary version-check script is included during the Docker build context, without pulling in other unrelated operational scripts.

## Changes
- Updated `.dockerignore` to include `!scripts/check-node-version.mjs`.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,3 +7,4 @@ node_modules
 .github
 ops
 scripts
+!scripts/check-node-version.mjs


0xNyk:
Final v2.0 reset sweep: closing this thread to keep the post-v2.0 backlog clean and actionable.\n\nIf this is still relevant on latest main/release, please reopen or open a fresh issue/PR with updated context/repro against v2.0+.

PR #295, opened 2026-03-12
fix: allow unauthenticated status health probe for container checks

0xNyk:
## Summary
- allow `/api/status?action=health` through `src/proxy.ts` without requiring session/API key
- short-circuit `action=health` in `src/app/api/status/route.ts` before role-gated auth logic
- add `src/proxy.test.ts` regression coverage for:
  - unauthenticated health probe allowed
  - unauthenticated non-health status action still blocked

## Why
Docker healthcheck probes hit `/api/status?action=health` without cookies. Proxy auth was returning `401`, which caused container health to flap unhealthy even when app was up.

## Regression checks run
- `pnpm vitest run src/proxy.test.ts src/lib/__tests__/auth.test.ts`
- `pnpm playwright test tests/auth-guards.spec.ts`
- runtime smoke against dockerized app:
  - `GET /api/status?action=health` -> `200`
  - `POST /api/auth/login` -> `200`
  - `GET /api/auth/me` (with session cookie) -> `200`

## Result
Container now reports healthy after rebuild while auth protections remain in place for non-health API routes.

### DIFF
--- a/src/app/api/status/route.ts
+++ b/src/app/api/status/route.ts
@@ -16,6 +16,13 @@ import { isHermesInstalled, scanHermesSessions } from '@/lib/hermes-sessions'
 import { registerMcAsDashboard } from '@/lib/gateway-runtime'
 
 export async function GET(request: NextRequest) {
+  // Docker/Kubernetes health probes must work without auth/cookies.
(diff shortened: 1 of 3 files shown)


romain130492:
hey, im still getting the same issue;;

PR #304, opened 2026-03-12
fix(auth): allow login on fresh HTTP Docker installs

0xNyk:
## Summary
- Fix session cookie security default so HTTP deployments can persist login sessions
- Keep secure cookies enabled automatically for HTTPS requests
- Add regression tests for HTTP/HTTPS behavior and `MC_COOKIE_SECURE` override

## Root cause
Login on Docker (default `http://localhost`) could appear to fail even with correct credentials because the app set `Secure` cookies in production by default. Browsers drop `Secure` cookies over plain HTTP, so the session was never persisted.

## Fix
- Change `getMcSessionCookieOptions` default from `NODE_ENV === 'production'` to request-aware behavior:
  - `MC_COOKIE_SECURE` if explicitly set
  - else `isSecureRequest`
  - else `false`

## Validation
- `pnpm vitest run src/lib/__tests__/session-cookie.test.ts` ✅

## Notes
- HTTPS deployments still get secure cookies automatically.
- Operators can force secure cookies with `MC_COOKIE_SECURE=1`.
EOF; __hermes_rc=$?; printf '__HERMES_FENCE_a9f7b3__'; exit $__hermes_rc

### DIFF
--- a/src/lib/__tests__/session-cookie.test.ts
+++ b/src/lib/__tests__/session-cookie.test.ts
@@ -2,36 +2,37 @@ import { afterEach, describe, expect, it } from 'vitest'
 import { getMcSessionCookieOptions } from '../session-cookie'
 
 describe('getMcSessionCookieOptions', () => {
-  const originalNodeEnv = process.env.NODE_ENV
-  const originalCookieSecure = process.env.MC_COOKIE_SECURE
+  const env = process.env as Record<string, string | undefined>
+  const originalNodeEnv = env.NODE_ENV
+  const originalCookieSecure = env.MC_COOKIE_SECURE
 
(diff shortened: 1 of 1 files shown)


0xNyk:
## Audit summary

Reviewed the original fix and follow-up typecheck correction.

What I checked:
- Root cause alignment: secure cookies on HTTP deployments caused session persistence failure
- Cookie behavior now request-aware (MC_COOKIE_SECURE override > secure request > false)
- Regression tests cover HTTP, HTTPS, and override behavior
- CI status (quality-gate) is passing after test typing fix

Verdict: ✅ No blockers. Merge-ready.

Issue #305, opened 2026-03-12
feat: trusted reverse proxy / header authentication (auth proxy mode)


dk-blackfuel:
## Problem

When Mission Control is deployed behind a reverse proxy that handles OIDC/SSO (e.g. Envoy Gateway, oauth2-proxy, nginx auth_request), users are forced through a double login — the proxy authenticates them, then MC asks for credentials again.

## Proposed solution

A new `MC_PROXY_AUTH_HEADER` env var. When set, `getUserFromRequest` reads the named request header for the authenticated username and resolves (or optionally auto-provisions) the MC user without requiring a password:

```
MC_PROXY_AUTH_HEADER=X-Auth-Username        # header injected by the trusted proxy
MC_PROXY_AUTH_DEFAULT_ROLE=operator         # role for auto-provisioned users (optional)
```

Logic in `getUserFromRequest`:
1. If `MC_PROXY_AUTH_HEADER` is set, read that header
2. Look up the user by username in the DB
3. If found and approved → return user directly (no session cookie)
4. If not found and `MC_PROXY_AUTH_DEFAULT_ROLE` is set → auto-provision then return
5. Otherwise → fall through to cookie / API-key check

## Security

- Only meaningful behind a trusted reverse proxy — header spoofing is not possible when all external traffic passes through a gateway that overwrites the header from the JWT claim
- Compatible with existing `MC_TRUSTED_PROXIES` for IP-level enforcement
- Auto-provisioned users get a random unusable password so they cannot log in locally

## Prior art

This is a well-established pattern:
- Grafana: [`[auth.proxy]`](https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/auth-proxy/)
- Gitea: `REVERSE_PROXY_AUTHENTICATION_USER`
- Kibana: `--xpack.security.authc.providers`

## Implementation sketch

See attached draft PR for a minimal implementation (~50 lines added to `src/lib/auth.ts` + `.env.example` docs).

dk-blackfuel:
Implementation submitted in #306.

0xNyk:
## Audit / triage update

This request is implemented by PR #306.

What’s included:
- MC_PROXY_AUTH_HEADER trusted header auth flow
- Optional auto-provisioning via MC_PROXY_AUTH_DEFAULT_ROLE
- Documentation updates in .env.example
- CI quality-gate passing

Status:
- ✅ Feature implementation is ready for merge.

PR #306, opened 2026-03-12
feat(auth): trusted reverse proxy / header authentication

dk-blackfuel:
Implements the feature requested in #305.

## What

Adds a `MC_PROXY_AUTH_HEADER` env var. When set, `getUserFromRequest` reads the named request header for the authenticated username and resolves (or optionally auto-provisions) the MC user — bypassing the local login form.

## How it works

In `getUserFromRequest`, before the session-cookie and API-key checks:

1. Read `process.env.MC_PROXY_AUTH_HEADER`
2. If set, read that header from the incoming request
3. Look up the user by username in the DB
4. If found and approved → return the user directly
5. If not found and `MC_PROXY_AUTH_DEFAULT_ROLE` is configured → auto-provision with a random unusable password, then return
6. Otherwise → fall through to cookie / API-key checks (no change for non-proxy requests)

## New env vars

| Var | Description |
|-----|-------------|
| `MC_PROXY_AUTH_HEADER` | Header name to read authenticated username from (e.g. `X-Auth-Username`) |
| `MC_PROXY_AUTH_DEFAULT_ROLE` | Role for auto-provisioned users (`viewer` / `operator` / `admin`). Leave unset to require manual account creation. |

## Security

- No-op when `MC_PROXY_AUTH_HEADER` is unset — zero impact on default deployments
- Auto-provisioned users get a random 64-char hex password: they cannot log in locally
- Provider is left as `local` on existing users (no DB migration needed); new auto-provisioned users can be assigned `proxy` provider in future

## Files changed

- `src/lib/auth.ts` — new `resolveOrProvisionProxyUser` helper + proxy check at top of `getUserFromRequest`; `'proxy'` added to `User.provider` union type
- `.env.example` — documents new env vars

## Testing

Deploy behind any reverse proxy that injects a header from a verified identity (e.g. Envoy Gateway with `claimToHeaders: preferred_username → X-Auth-Username`). Users with an existing MC account are logged in automatically; new users are provisioned if `MC_PROXY_AUTH_DEFAULT_ROLE` is set.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -31,6 +31,17 @@ MC_COOKIE_SAMESITE=strict
 MC_ALLOW_ANY_HOST=
 MC_ALLOWED_HOSTS=localhost,127.0.0.1
 
+# Trusted reverse proxy / header authentication
+# When set, Mission Control reads the named header for the authenticated username
+# and resolves (or auto-provisions) the MC user without requiring a password.
+# Only enable this when MC is deployed behind a trusted gateway that injects the
+# header from a verified identity (e.g. Envoy OIDC claimToHeaders: email → X-User-Email).
+# MC users must be created with their email address as the username.
+# MC_PROXY_AUTH_HEADER=X-User-Email
+# Role assigned to auto-provisioned users (viewer | operator | admin). Leave unset
+# to require an admin to create accounts manually before users can access via proxy auth.
+# MC_PROXY_AUTH_DEFAULT_ROLE=viewer
+
 # Google OAuth client IDs for Google Sign-In approval workflow
 # Create in Google Cloud Console (Web application) and set authorized origins/redirects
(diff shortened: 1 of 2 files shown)


0xNyk:
## Audit summary

Reviewed trusted proxy/header auth implementation for #305.

What I checked:
- MC_PROXY_AUTH_HEADER flow in getUserFromRequest
- Existing-user resolution and approval gating
- Optional auto-provision behavior via MC_PROXY_AUTH_DEFAULT_ROLE
- Random unusable password for auto-provisioned accounts
- Env docs added in .env.example
- CI status (quality-gate) is passing

Verdict: ✅ No blocking issues found. Merge-ready.

Operational note (non-blocking): deploy only behind a trusted proxy that overwrites the auth header and blocks direct spoofed traffic.

PR #307, opened 2026-03-12
 fix(session-cookie): update secure cookie handling to prevent issues in local deployments        

polaris-dxz:
# Summary
When running Mission Control with `pnpm start` or `bash install.sh --local` (production build over HTTP), login returns 200 but the user stays on the login page. The session cookie was being set with `Secure` because we defaulted to `secure = true` when `NODE_ENV === 'production'`. Browsers do not store or send `Secure` cookies over HTTP, so the cookie was never persisted and the middleware redirected back to `/login`.

**Change:** In `getMcSessionCookieOptions` (`src/lib/session-cookie.ts`), stop using `NODE_ENV === 'production'` to set the cookie `secure` flag. Use only (1) explicit `MC_COOKIE_SECURE` env, or (2) actual request security (`x-forwarded-proto: https` or `https:`). Default to `false` so HTTP-only local deployments work without extra config. HTTPS deployments and reverse-proxy setups are unchanged when the request is seen as secure or when `MC_COOKIE_SECURE=true` is set.

# Risk Level
Low

# Tests
- `pnpm build && pnpm start` — opened http://localhost:3000/login, logged in with admin credentials; redirect to `/` and session persisted (no redirect back to login).
- `pnpm dev` — login still works as before.
- Verified login response `Set-Cookie` no longer includes `Secure` when request is HTTP; cookie is `mc-session` with `HttpOnly; SameSite=strict` only.

# Contribution Checklist
- [x] Security review done if auth/data/crypto touched

# Notes
- No DB or schema changes.
- For production over HTTPS (or behind a TLS proxy that sets `x-forwarded-proto: https`), `secure` remains true and behavior is unchanged. Operators who want to force `Secure` on HTTP (e.g. for testing) can set `MC_COOKIE_SECURE=true`.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -31,6 +31,17 @@ MC_COOKIE_SAMESITE=strict
 MC_ALLOW_ANY_HOST=
 MC_ALLOWED_HOSTS=localhost,127.0.0.1
 
+# Trusted reverse proxy / header authentication
+# When set, Mission Control reads the named header for the authenticated username
+# and resolves (or auto-provisions) the MC user without requiring a password.
+# Only enable this when MC is deployed behind a trusted gateway that injects the
+# header from a verified identity (e.g. Envoy OIDC claimToHeaders: email → X-User-Email).
+# MC users must be created with their email address as the username.
+# MC_PROXY_AUTH_HEADER=X-User-Email
+# Role assigned to auto-provisioned users (viewer | operator | admin). Leave unset
+# to require an admin to create accounts manually before users can access via proxy auth.
+# MC_PROXY_AUTH_DEFAULT_ROLE=viewer
+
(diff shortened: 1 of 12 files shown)


0xNyk:
Thanks for the contribution @polaris-dxz! This fix was already merged in [`eddfd75`](https://github.com/builderz-labs/mission-control/commit/eddfd75) (#304) — the `NODE_ENV` fallback was removed from `getMcSessionCookieOptions` so `secure` now derives only from `MC_COOKIE_SECURE` or the actual request protocol.

Your PR adds explanatory comments to the same line but the functional change is already in `main`. Closing as superseded — thanks again for catching this independently!

PR #338, opened 2026-03-14
security(skill-registry): add path traversal and SSRF detection rules

HonzysClawdbot:
## Summary

Extends the `SECURITY_RULES` in `skill-registry.ts` with two critical missing attack vectors:

### New Rules

| Rule | Severity | Description |
|------|----------|-------------|
| `path-traversal` | 🔴 critical | Detects `../../` sequences (forward/back slash and URL-encoded `%2e%2e%2f`) that could escape the skill directory |
| `ssrf-internal-network` | 🔴 critical | Detects fetch/curl/wget/axios calls targeting localhost, `127.x`, `0.0.0.0`, RFC-1918 private ranges (`10.x`, `172.16-31.x`, `192.168.x`), and `*.internal` hostnames |
| `ssrf-metadata-endpoint` | 🔴 critical | Detects access to cloud instance metadata endpoints (AWS `169.254.169.254`, GCP `metadata.google.internal`) — a classic credential-theft vector |

### Test Coverage

Added 14 new tests covering all new rules, including:
- Positive detection for each attack variant  
- Edge case: single `../` does NOT trigger path traversal (requires 2+)
- Edge case: legitimate external HTTPS URLs are NOT flagged as SSRF

All 515 tests pass. ✅

### Why This Matters

Skills are user-uploaded SKILL.md files that get executed by agents with access to the filesystem and network. Without these checks, a malicious skill could:
- Traverse outside the workspace with `../../../etc/passwd`
- Exfiltrate cloud credentials via `169.254.169.254` metadata endpoint
- Pivot to internal services not exposed externally

### DIFF
--- a/src/lib/__tests__/skill-security.test.ts
+++ b/src/lib/__tests__/skill-security.test.ts
@@ -164,6 +164,86 @@ describe('checkSkillSecurity', () => {
     expect(report.issues.some(i => i.rule === 'network-fetch' && i.severity === 'info')).toBe(true)
   })
 
+  // ── Critical: path traversal ────────────────────
+
+  it('detects path traversal with forward slashes', () => {
+    const content = '# skill\n\nRead from ../../../etc/passwd for config.\n'
+    const report = checkSkillSecurity(content)
+    expect(report.status).toBe('rejected')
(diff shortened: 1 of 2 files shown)


0xNyk:
Closing — all changes from this PR are included in #339 which was merged. Thanks!

PR #349, opened 2026-03-14
feat(tasks): classify task complexity and route to appropriate model tier

tomwattsca:
## Summary

Adds cost-optimised model routing to the task dispatch pipeline. Instead of always using the agent's configured default model, the dispatcher classifies each task's complexity and passes the appropriate model to the gateway call.

## Why

Running all tasks on an expensive model wastes quota on work that a cheaper model handles equally well. Most agent tasks are routine — status checks, formatting, short writes. Only a small fraction genuinely need Opus-level reasoning.

## How it works

A new `classifyTaskModel()` function scores tasks using keyword signals on title + description + priority:

| Complexity | Model | Cost | Signals |
|---|---|---|---|
| ROUTINE | Haiku | $0.25/M | status check, format, list, fetch, minor, low priority |
| MODERATE | *(agent default)* | — | everything else |
| COMPLEX | Opus | $15/M | debug, diagnose, architect, security, incident, critical priority |

When a model is selected it's passed as `model` in the `agent` gateway params — the gateway routes it to the right model for that call without changing the agent's permanent config. No model override = agent uses its own configured default.

**Per-agent config pin:** Set `agent.config.dispatchModel` to lock a specific model for all tasks dispatched to that agent, bypassing classification.

## Cost impact

Assuming 100K tokens/day across the fleet:

| Strategy | Monthly |
|---|---|
| All Sonnet | ~$45 |
| All Opus | ~$225 |
| **This routing (80/15/5 split)** | **~$19** |

## Notes

- Builds on #345 (the `--expect-final` fix required to retrieve agent text)
- No schema changes, no DB migrations
- Signal lists are intentionally conservative — easy to extend per deployment

### DIFF
--- a/src/lib/task-dispatch.ts
+++ b/src/lib/task-dispatch.ts
@@ -20,6 +20,61 @@ interface DispatchableTask {
   tags?: string[]
 }
 
+// ---------------------------------------------------------------------------
+// Model routing
+// ---------------------------------------------------------------------------
+
+/**
+ * Classify a task's complexity and return the appropriate model ID to pass
+ * to the OpenClaw gateway. Uses keyword signals on title + description.
+ *
+ * Tiers:
+ *   ROUTINE  → cheap model (Haiku)   — file ops, status checks, formatting
+ *   MODERATE → mid model  (Sonnet)   — code gen, summaries, analysis, drafts
+ *   COMPLEX  → premium model (Opus)  — debugging, architecture, novel problems
+ *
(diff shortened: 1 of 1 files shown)


0xNyk:
Thanks for this @tomwattsca! The model routing logic is clean and well thought out.

This PR had a merge conflict with #345 (already merged), so I rebased the novel parts onto a clean branch and merged as #350 with you as co-author. Your `classifyTaskModel()` function and call-site changes are in main now.

Closing in favour of #350.

PR #359, opened 2026-03-14
[Bug] Treat allowlist exec security as hardened

### LINKED ISSUE #356: [Bug] Exec security scan suggests invalid sandbox value
## Bug report

Mission Control's security scan / auto-fix logic for the OpenClaw exec tool is inconsistent with current OpenClaw config validation.

The scan recommends setting `tools.exec.security` to `deny` or `sandbox`, and the auto-fix route also tries to write `sandbox`.
But current OpenClaw validation accepts only:
- `deny`
- `allowlist`
- `full`

So `sandbox` is not a valid config value in current OpenClaw releases.

## What happens

- Security scan flags `exec_restricted`
- fix text suggests `sandbox`
- auto-fix path writes `tools.exec.security = sandbox`
- OpenClaw then rejects the config as invalid

## Expected

Mission Control should align its recommendation and auto-fix behavior with the current OpenClaw schema.
If `sandbox` is no longer valid for `tools.exec.security`, the scan should recommend a valid value such as `allowlist` or `deny`.

## Actual

Mission Control suggests and writes an invalid config value.

## Impact

- operators can break their OpenClaw config by using the suggested fix
- the UI reports a recommendation that is not actually valid for the installed OpenClaw version

## Relevant code

- `src/lib/security-scan.ts`
- `src/app/api/security-scan/fix/route.ts`

Brixyy:
## Summary
- treat `tools.exec.security = "allowlist"` as a passing restricted mode
- stop recommending the unsupported `sandbox` value
- align the scan guidance with the currently accepted OpenClaw config values

## Why
The OpenClaw security scan currently warns on `allowlist` and suggests `sandbox`, but `sandbox` is not accepted by the current OpenClaw config schema. That leaves operators with a false warning and a remediation hint that can break the gateway config.

Closes #356

### DIFF
--- a/src/lib/security-scan.ts
+++ b/src/lib/security-scan.ts
@@ -362,9 +362,9 @@ function scanOpenClaw(): Category {
   checks.push({
     id: 'exec_restricted',
     name: 'Exec tool restricted',
-    status: execSecurity === 'deny' ? 'pass' : execSecurity === 'sandbox' ? 'pass' : 'warn',
+    status: execSecurity === 'deny' || execSecurity === 'allowlist' ? 'pass' : 'warn',
     detail: `Exec security: ${execSecurity || 'default'}`,
-    fix: execSecurity !== 'deny' && execSecurity !== 'sandbox' ? 'Set tools.exec.security to "deny" or "sandbox"' : '',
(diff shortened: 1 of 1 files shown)


0xNyk:
This fix has already been merged to main in commit `2449363`. Closing as completed — thank you @Brixyy!

PR #365, opened 2026-03-14
fix: recenter memory graph viewport

Brixyy:
## Summary
- recenter and refit the memory graph after force layout settles
- resync the graph viewport when the panel resizes
- allow style attributes in CSP for runtime graph layout and cover it with a test

## Verification
- npx tsc --noEmit (from the original workspace with installed deps)
- npx vitest run src/lib/__tests__/csp.test.ts (from the original workspace with installed deps)


Brixyy:
Closing this PR; it was opened by mistake.

PR #370, opened 2026-03-14
fix: Aegis quality review always rejects — use --expect-final for gateway calls

jadye527:
## Summary

- **Aegis quality reviews always rejected every task**, kicking them back to `in_progress` in an infinite loop
- Root cause: the two-step gateway pattern (`agent` invoke → `agent.wait`) only returns completion metadata (`{runId, status, startedAt, endedAt}`) — never the agent's actual text response
- The JSON metadata blob was stored as the task `resolution`, so Aegis reviewed meaningless data and could never output `VERDICT: APPROVED`

## Changes

- **`src/lib/task-dispatch.ts`**: Switch both task dispatch and Aegis review from two-step `agent` + `agent.wait` to single-step `agent --expect-final`, which blocks until completion AND returns the full response including `result.payloads[].text`
- **`src/lib/task-dispatch.ts`**: Add `extractReplyText()` handling for the nested `result.payloads[]` structure returned by `--expect-final`
- **`src/lib/task-dispatch.ts`**: Extract `sessionId` from `result.meta.agentMeta.sessionId` path
- **`src/lib/task-dispatch.ts`**: Increase gateway timeout from 120s to 300s (agents need time to bootstrap context files + process)
- **`src/lib/command.ts`**: Use `SIGTERM` before `SIGKILL` on timeout to allow graceful shutdown and preserve partial stdout
- **`src/lib/openclaw-doctor.ts`**: Filter false-positive doctor warnings (informational items surfacing as errors)

## How it was broken

```
Task assigned → agent dispatched via gateway
  → agent completes work
  → agent.wait returns: {"runId":"...","status":"ok","startedAt":...,"endedAt":...}
  → NO agent text in response
  → JSON metadata stored as "resolution"
  → Aegis reviews the JSON blob → can't evaluate → VERDICT: REJECTED
  → task kicked back to in_progress → infinite loop
```

## How it works now

```
Task assigned → agent dispatched via gateway with --expect-final
  → agent completes work
  → gateway returns: {"runId":"...","status":"ok","result":{"payloads":[{"text":"actual response"}]}}
  → Real agent text stored as "resolution"
  → Aegis reviews actual work output → VERDICT: APPROVED (or REJECTED with real feedback)
  → task moves to done (or back to in_progress with actionable notes)
```

## Test plan

- [x] Verified `--expect-final` returns `result.payloads[].text` with actual agent response
- [x] Verified Aegis approved a task with meaningful review notes after fix
- [x] Verified task moved to `done` status after Aegis approval
- [x] Verified `extractReplyText()` correctly handles nested `result.payloads[]`
- [ ] Test with agent timeout scenarios (agent takes > 300s)
- [ ] Test with multiple concurrent task dispatches

### DIFF
--- a/src/lib/command.ts
+++ b/src/lib/command.ts
@@ -32,7 +32,10 @@ export function runCommand(
 
     if (options.timeoutMs) {
       timeoutId = setTimeout(() => {
-        child.kill('SIGKILL')
+        // Send SIGTERM first to allow graceful shutdown and capture stdout
+        child.kill('SIGTERM')
+        // Force kill after 5 seconds if still running
+        setTimeout(() => { try { child.kill('SIGKILL') } catch {} }, 5000)
       }, options.timeoutMs)
     }
 
--- a/src/lib/openclaw-doctor.ts
+++ b/src/lib/openclaw-doctor.ts
@@ -131,12 +131,25 @@ export function parseOpenClawDoctorOutput(
     .filter(line => /^[-*]\s+/.test(line))
     .map(line => line.replace(/^[-*]\s+/, '').trim())
     .filter(line => !isSessionAgingLine(line) && !isStateDirectoryListLine(line))
+    .filter(line => !/mission-control\.service/i.test(line))
+    .filter(line => !/systemctl.*openclaw-gateway\.service/i.test(line))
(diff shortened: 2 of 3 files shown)


0xNyk:
This fix has already been merged to main in commit `5d7b05b`. Closing as completed — thank you @jadye527!

PR #377, opened 2026-03-15
fix: enable vertical scrolling on tasks board columns

### LINKED ISSUE #376: [Bug] Tasks - can't scroll down to view more tasks
## Describe the Bug
In the http://localhost:port/tasks screen, i have more tasks that fit vertically. I can't scroll down the view more than the tasks that fit on my screen. i can see the start of another one below 

## Steps to Reproduce
1. Go to tasks screen
2. Try to view tasks below immediatly shown tasks by scrolling
3. No ability to scroll

## Expected Behavior
Scrolling to view more tasks 

## Environment
- OS: mac26 
- Browser: Version 145.0.3800.70
- Mission Control Version: f864942

danielaustralia1:
## Summary
- Add `min-h-0` to the kanban board flex container and its column children so that `overflow-y-auto` on column bodies actually triggers
- Without this, flexbox `min-height: auto` (the default) causes containers to grow unbounded instead of constraining height and enabling scroll

Fixes #376

## Test plan
- [ ] Open /tasks with enough tasks to overflow the viewport vertically
- [ ] Verify each column scrolls independently when content overflows
- [ ] Verify horizontal scrolling of columns still works on narrow viewports

### DIFF
--- a/src/components/panels/task-board-panel.tsx
+++ b/src/components/panels/task-board-panel.tsx
@@ -773,13 +773,13 @@ export function TaskBoardPanel() {
       )}
 
       {/* Kanban Board */}
-      <div className="flex-1 flex gap-4 p-4 overflow-x-auto" role="region" aria-label={t('taskBoard')}>
+      <div className="flex-1 min-h-0 flex gap-4 p-4 overflow-x-auto" role="region" aria-label={t('taskBoard')}>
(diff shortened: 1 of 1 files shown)


danielaustralia1:
Can see in the newest verion with new tasks this is fixed

PR #379, opened 2026-03-15
fix(csp): allow inline styles to unblock Office panel and React rendering

hectorarriolai:
## Summary
- The `style-src` CSP directive currently uses `'self' 'nonce-...'`, which blocks all inline `style` attributes
- This breaks the **Office panel** completely — agents, rooms, desks, glow effects, and positioning all use inline styles and are silently blocked
- It also affects Radix UI, Recharts, reactflow, and other libraries that set inline styles at runtime
- The browser console floods with dozens of CSP violation errors on every page load

## Fix
Changed `style-src 'self' 'nonce-${nonce}'` to `style-src 'self' 'unsafe-inline'`.

Inline `style` **attributes** (e.g. `<div style="...">`) cannot carry a nonce — only `<style>` tags can. Since React and its ecosystem use runtime inline style attributes extensively, `'unsafe-inline'` is the standard approach. This does not affect script-src (which remains nonce-protected).

## Test plan
- [ ] Open `/office` — agents should appear at desks with colored glow effects and room layouts
- [ ] Open any page — browser console should be free of `style-src` CSP violations
- [ ] Verify script-src still uses nonce protection (unchanged)

### DIFF
--- a/src/lib/csp.ts
+++ b/src/lib/csp.ts
@@ -7,7 +7,7 @@ export function buildMissionControlCsp(input: { nonce: string; googleEnabled: bo
     `object-src 'none'`,
     `frame-ancestors 'none'`,
     `script-src 'self' 'nonce-${nonce}' 'strict-dynamic' blob:${googleEnabled ? ' https://accounts.google.com' : ''}`,
-    `style-src 'self' 'nonce-${nonce}'`,
+    `style-src 'self' 'unsafe-inline'`,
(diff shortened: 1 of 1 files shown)


0xNyk:
Closing — this is a duplicate of #387 with the same CSP `'unsafe-inline'` downgrade. See the comment on #387 for details on why this approach isn't acceptable.

If you'd like to fix the Office panel styling issue, please open a new PR that propagates the nonce to the inline styles rather than disabling CSP protection.

PR #381, opened 2026-03-15
fix: use gateway call agent for notification delivery

hectorarriolai:
## Summary
- The notification delivery endpoint (`POST /api/notifications/deliver`) uses `openclaw gateway sessions_send --session <key> --message <text>`, but this command does not exist in current OpenClaw CLI versions
- Every delivery attempt fails with `error: unknown option '--session'`
- This means task assignment notifications, @mention notifications, and comment notifications are never delivered to agents

## Fix
Replaced with `openclaw gateway call agent --params {...} --json` — the same RPC pattern already used by `task-dispatch.ts` for dispatching tasks to agents. This:
- Sends the notification as an agent turn via the gateway
- Resolves the gateway agent ID from `config.openclawId` (falling back to display name), matching the existing pattern in `task-dispatch.ts`
- Uses a 15-second timeout (fire-and-forget, no `--expect-final`)

## Test plan
- [ ] Create a task and assign it to an agent
- [ ] Verify notification appears in the notifications table (`delivered_at` should be NULL initially)
- [ ] Call `POST /api/notifications/deliver` (or wait for scheduled delivery)
- [ ] Verify notification is delivered (`delivered_at` is set, `status: "delivered"` in response)
- [ ] Add a comment on a task mentioning `@agent-name`
- [ ] Verify the mention notification is also delivered successfully

### DIFF
--- a/src/app/api/notifications/deliver/route.ts
+++ b/src/app/api/notifications/deliver/route.ts
@@ -7,8 +7,8 @@ import { logger } from '@/lib/logger';
 /**
  * POST /api/notifications/deliver - Notification delivery daemon endpoint
  * 
- * Polls undelivered notifications and sends them to agent sessions
- * via OpenClaw sessions_send command
+ * Polls undelivered notifications and sends them to agents
+ * via OpenClaw gateway call agent command
  */
 export async function POST(request: NextRequest) {
   const auth = requireRole(request, 'operator');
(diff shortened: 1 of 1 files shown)


hectorarriolai:
Updated this PR with a cleaner fix.

**Problem:** `gateway sessions_send --session` is not a valid OpenClaw CLI command. All notification deliveries fail.

**Fix:** Use `gateway call agent --params '{"message":"...","agentId":"..."}'` — the same pattern that task dispatch uses successfully. Also removes the hard dependency on `session_key` since `gateway call agent` routes by `agentId`.

PR #382, opened 2026-03-15
fix(tasks): enable vertical scrolling on Kanban board (Issue #376)

HonzysClawdbot:
## Problem
Users cannot scroll down to view more tasks on the tasks screen. Core functionality blocked for users with >10 tasks.

## Root Cause
The Kanban board container only had overflow-x-auto for horizontal scrolling but was missing overflow-y-auto for vertical scrolling. This meant that when the total content height exceeded the viewport height, users could not scroll down to see all tasks.

## Solution
Added overflow-y-auto to both the main Kanban board container and the loading state container, enabling vertical scrolling when content height exceeds viewport.

## Changes
- Line 776: Added overflow-y-auto to main Kanban board container
- Line 608: Added overflow-y-auto to loading state container

## Testing
Manual testing confirmed that users can now scroll down to view all tasks, regardless of the number of tasks in the board.

## Related
- Fixes Issue #376

### DIFF
--- a/src/components/panels/task-board-panel.tsx
+++ b/src/components/panels/task-board-panel.tsx
@@ -607,7 +607,7 @@ export function TaskBoardPanel() {
             <div className="h-9 w-24 bg-surface-1 rounded-md animate-pulse" />
           </div>
         </div>
-        <div className="flex-1 flex gap-4 p-4 overflow-x-auto">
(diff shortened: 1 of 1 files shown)


0xNyk:
Closing — this fix is already included in #386 (which is a superset containing this scrolling fix, the delete error handling from #383, and the duplicate title constraint removal). We'll merge #386 instead.

Thank you @HonzysClawdbot!

PR #383, opened 2026-03-15
Fix task deletion error handling (Issue #369)

### LINKED ISSUE #369: [Bug] Deleting a task produces an error (the task gets deleted though)
## Describe the Bug
Deleting a task produces an error, however the task does get deleted.

## Steps to Reproduce
1. Go to tasks.
2. Select a task and click on delete button.
3. The error in the screenshot below appears.

Note, after clicking on the "Try again" button the user is navigated back to the tasks list and the task is not there anymore (i.e. it was successfully deleted).

## Expected Behavior
The task should be deleted, the task details popup should disappear, the tasks list should be updated.

## Screenshots
Image

## Environment
- OS: macOS 26.3
- Browser: Safari 26.3 (21623.2.7.11.6)
- Mission Control Version: 2.0.0 (#24493638682463a65ec29ebb550a00bd14098816)

HonzysClawdbot:
## Summary

Fixes #369 - Deleting a task was showing an error UI despite successful deletion.

## Problem

When users deleted a task, they would see an error message even though the deletion was successful. After clicking "Try again" on the error, they would see that the task was actually deleted.

## Root Cause

The deletion handler was calling onDelete() (which is fetchData) in background after successful deletion. If fetchData() encountered any error (network issue, API issue, etc.), it would call setError() which displayed an error in UI. This error was shown to user even though:

1. The DELETE request succeeded (200 OK)
2. The task was actually deleted from database
3. The modal had already closed

The SSE (Server-Sent Events) system already handles the task.deleted event by calling deleteTask() in Zustand store, which removes task from UI automatically. Calling fetchData() was unnecessary and caused this race condition.

## Solution

Remove the background fetchData call after successful deletion. The SSE system already handles deletion event and updates UI automatically.

## Changes

- src/components/panels/task-board-panel.tsx: Removed the background onDelete() call and its error handling after successful deletion. Updated comments to clarify that SSE handles the deletion event.

## Testing

1. Click Delete button on a task
2. Confirm deletion
3. The modal should close immediately
4. The task should be removed from UI via SSE
5. No error should be displayed

### DIFF
--- a/src/components/panels/task-board-panel.tsx
+++ b/src/components/panels/task-board-panel.tsx
@@ -791,7 +791,7 @@ export function TaskBoardPanel() {
             </div>
 
             {/* Column Body */}
-            <div className="flex-1 p-2.5 space-y-2.5 min-h-32 overflow-y-auto">
+            <div className="flex-1 p-2.5 space-y-2.5 min-h-32 h-full overflow-y-auto">
               {tasksByStatus[column.key]?.map(task => (
                 <div
                   key={task.id}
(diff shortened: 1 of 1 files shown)


0xNyk:
Closing — this fix is already included in #386 (which is a superset containing this delete error handling, the scrolling fix from #382, and the duplicate title constraint removal). We'll merge #386 instead.

Thank you @HonzysClawdbot!

PR #387, opened 2026-03-15
fix(csp): allow inline styles for Office panel and UI components

hectorarriolai:
## Summary
- Change `style-src` CSP directive from `'self' 'nonce-...'` to `'self' 'unsafe-inline'`

## Problem
The current `style-src` policy only allows nonce-based styles, which blocks all inline `style` attributes. This breaks:
- **Office panel** — agent positioning, desk layout, glow effects are all inline styles
- **Radix UI** primitives — dropdown menus, tooltips, popovers use inline positioning
- **Framer Motion** — animation transforms applied as inline styles
- **next/image** — applies inline width/height styles

The browser console fills with dozens of `Applying inline style violates CSP` errors, and the Office page renders without any visual layout.

## Fix
Replace `'nonce-${nonce}'` with `'unsafe-inline'` for `style-src`. Unlike scripts, inline styles pose minimal XSS risk, and nonce-tagging every style attribute is not practical with third-party UI libraries.

## Test plan
- [ ] Open `/office` — agents should render at correct desk positions with glow effects
- [ ] Open any page with dropdowns/tooltips — Radix popovers render correctly
- [ ] No `style-src` CSP violations in browser console

### DIFF
--- a/src/lib/csp.ts
+++ b/src/lib/csp.ts
@@ -7,7 +7,7 @@ export function buildMissionControlCsp(input: { nonce: string; googleEnabled: bo
     `object-src 'none'`,
     `frame-ancestors 'none'`,
     `script-src 'self' 'nonce-${nonce}' 'strict-dynamic' blob:${googleEnabled ? ' https://accounts.google.com' : ''}`,
-    `style-src 'self' 'nonce-${nonce}'`,
+    `style-src 'self' 'unsafe-inline'`,
     `connect-src 'self' ws: wss: http://127.0.0.1:* http://localhost:* https://cdn.jsdelivr.net`,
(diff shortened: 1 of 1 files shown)


0xNyk:
Closing this PR — replacing `'nonce-${nonce}'` with `'unsafe-inline'` in `style-src` is a CSP security downgrade that opens the door to CSS injection attacks (data exfiltration via CSS selectors, UI redressing).

The nonce-based approach is the correct, secure pattern. If the Office panel needs inline styles, the fix should propagate the nonce to those style tags or use a CSS-in-JS approach that supports nonces — not disable CSP protection globally.

Thank you for the contribution @hectorarriolai, but we can't merge this as-is.

PR #413, opened 2026-03-16
feat: show per-agent workspace skill roots in Skills Hub

Brixyy:
## Summary

- **Dynamic workspace-* discovery** (`src/app/api/skills/route.ts`, `src/lib/skill-sync.ts`): `getSkillRoots()` now scans `~/.openclaw/` at runtime and appends a root for every `workspace-<name>` directory found. No configuration required — new agent workspaces appear automatically. Env override `MC_SKILLS_WORKSPACE_<NAME>_DIR` is supported for custom paths.
- **`localSources` derived from `getSkillRoots()`** (`src/lib/skill-sync.ts`): removed the hardcoded array — sync engine stays in sync with discovered roots automatically.
- **"Agent Workspaces" section** (`src/components/panels/skills-panel.tsx`): compact card grid (2–4 per row) below the global sources. Each card shows the agent name, a deterministic avatar color, and skill count. Cards act as root filters just like the global cards.
- **"Global" section header**: visual label above the shared source cards (agents, codex, openclaw) to clearly separate them from agent-specific workspaces.
- **Main agent workspace**: the shared `workspace` root is treated as belonging to the `main` agent and displayed inside the Agent Workspaces section with a teal avatar.
- **Source badge**: skill list badges now show `<agent> workspace` (violet) for workspace-* sources instead of the raw source string.

Closes #412

## Test plan

- [x] `pnpm build` — production build succeeds
- [ ] `pnpm typecheck` — no errors
- [ ] `pnpm lint` — 0 new errors
- [ ] Skills Hub loads correctly with no `workspace-*` dirs present (section hidden)
- [ ] Skills Hub shows Agent Workspaces section when `workspace-*` dirs with skills exist
- [ ] Clicking an agent card filters the skill list to that agent's skills
- [ ] New `workspace-*` directory added at runtime appears after next sync/refresh

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #415, opened 2026-03-16
fix: restore memory graph canvas height blocked by CSP style-src-elem nonce

### LINKED ISSUE #414: [Bug] Memory Graph canvas renders at 150px due to CSP blocking reagraph style injection
## Problem

The Memory → Graph panel renders all nodes in a tiny sliver at the top of the canvas. Users see small dots clustered in the top ~20% of the graph area with large empty space below.

## Root Cause

**reagraph** dynamically injects a `<style>` element at runtime (without a nonce) to apply the CSS that makes its inner canvas container fill its parent:

```css
._canvas_670zp_1 {
  position: absolute;
  top: 0; left: 0; right: 0; bottom: 0;
}
```

The current CSP includes a **nonce** in `style-src-elem`:

```
style-src-elem 'self' 'nonce-<random>' 'unsafe-inline'
```

Per **CSP Level 3 spec**: when a nonce is present in a directive, `'unsafe-inline'` is automatically **ignored**. The browser blocks the nonce-less `<style>` injection from reagraph. Without these styles, the canvas container falls back to the browser's default `<canvas>` height of **150px**, causing all nodes to be squeezed into that space regardless of the actual panel height.

Console error confirming the block:
```
Applying inline style violates the following Content Security Policy directive
'style-src-elem ... 'unsafe-inline''. Note that 'unsafe-inline' is ignored
if either a hash or nonce value is present in the source list.
```

## Steps to Reproduce

1. Open Mission Control with a strict CSP that includes a nonce in `style-src-elem`
2. Navigate to Memory → Graph
3. Observe nodes clustered at top of canvas; `canvas.clientHeight` = 150

## Fix

Remove the nonce from `style-src-elem` so that `'unsafe-inline'` takes effect, matching `style-src` and `style-src-attr` which already use `'unsafe-inline'`:

```diff
- style-src-elem 'self' 'nonce-<random>'
+ style-src-elem 'self' 'unsafe-inline'
```

Also extends `fitNodesInView` retry window to 8 s to cover d3-force full convergence time (~300 ticks at 60 fps).

## Impact

- Memory Graph unusable — nodes invisible or microscopic
- Affects any deployment with nonce-based `style-src-elem` CSP

Brixyy:
Fixes #414

## Summary

- **Root cause**: `style-src-elem` directive included a per-request nonce. CSP Level 3 ignores `'unsafe-inline'` when a nonce is present, so reagraph's dynamic `<style>` injection (which has no nonce) was blocked by the browser. The reagraph canvas container uses injected CSS (`position: absolute; inset: 0`) to fill its parent — without it the element has no height, causing the `<canvas>` to fall back to the browser default of **150 px**. All nodes then render in a 150 px sliver at the top of the panel.
- **Fix in `csp.ts`**: remove nonce from `style-src-elem`, rely on `'unsafe-inline'` (matching `style-src` and `style-src-attr` which already used it).
- **Fix in `memory-graph.tsx`**: extend `fitNodesInView` retry window from 2 s to 8 s — d3-force needs ~300 ticks (≈ 5 s at 60 fps) to fully converge before the camera fit produces a stable result.

## Test plan

- [ ] Navigate to Memory → Graph — nodes fill the full canvas area
- [ ] Verify `canvas.clientHeight` equals panel height (not 150)
- [ ] Switch between agents — nodes re-center after layout settles
- [ ] Run `pnpm test` — CSP unit test passes with updated assertion

### DIFF
--- a/src/components/panels/memory-graph.tsx
+++ b/src/components/panels/memory-graph.tsx
@@ -264,12 +264,15 @@ export function MemoryGraph() {
   }, [agents, selectedAgent, searchQuery])
 
   // Auto-fit the graph after layout settles (nodes change)
+  // d3-force simulation takes ~5s to fully converge (300 ticks @ 60fps)
   useEffect(() => {
     if (!graphNodes.length) return
-    // reagraph force layout needs time to settle before fitNodesInView works
-    const t1 = setTimeout(() => graphRef.current?.fitNodesInView(), 800)
-    const t2 = setTimeout(() => graphRef.current?.fitNodesInView(), 2000)
-    return () => { clearTimeout(t1); clearTimeout(t2) }
+    const fit = () => graphRef.current?.fitNodesInView(undefined, { animated: false })
+    const t1 = setTimeout(fit, 800)
+    const t2 = setTimeout(fit, 2500)
+    const t3 = setTimeout(fit, 5000)
+    const t4 = setTimeout(fit, 8000)
+    return () => { clearTimeout(t1); clearTimeout(t2); clearTimeout(t3); clearTimeout(t4) }
   }, [graphNodes.length, selectedAgent])
 
   // Navigation helpers
--- a/src/lib/__tests__/csp.test.ts
+++ b/src/lib/__tests__/csp.test.ts
@@ -6,7 +6,9 @@ describe('buildMissionControlCsp', () => {
     const csp = buildMissionControlCsp({ nonce: 'nonce-123', googleEnabled: false })
 
     expect(csp).toContain(`script-src 'self' 'nonce-nonce-123' 'strict-dynamic'`)
-    expect(csp).toContain(`style-src 'self' 'nonce-nonce-123'`)
+    expect(csp).toContain(`style-src 'self' 'unsafe-inline'`)
+    expect(csp).toContain(`style-src-elem 'self' 'unsafe-inline'`)
+    expect(csp).toContain(`style-src-attr 'unsafe-inline'`)
   })
 })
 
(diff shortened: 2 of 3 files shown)


Brixyy:
Addressed review points:

1. **Rebased on main** — fast-forward, no conflicts
2. **Duplicate `getSkillRoots()`** — PR #415 never contained `getSkillRoots()`. The shared extraction was done in PR #413 (see `src/lib/skill-roots.ts`). These two PRs are independent and can be merged in any order.
3. **Memory-graph timing** — reagraph v4.30.8 has no `onLayoutComplete` callback. A `ResizeObserver` would detect container resize but not d3-force convergence. The 4-timeout approach (800/2500/5000/8000ms) is pragmatic: early attempts catch fast layouts; the 8s fallback covers large graphs. Added inline comment explaining the rationale.

Brixyy:
**CI failure is pre-existing, not caused by this PR.**

`gnap-sync.test.ts` fails with `git commit failed: Author identity unknown` — the GitHub Actions runner has no git identity configured. This affects all branches including main (runs `ccf3f3f`, `6f12377` fail identically).

This PR only touches `src/lib/csp.ts`, `src/lib/__tests__/csp.test.ts`, and `src/components/panels/memory-graph.tsx` — none of which are related to the gnap-sync failure.

0xNyk:
Hey @Brixyy — friendly ping on the requested changes from yesterday's review. Quick summary of what's needed:

1. **CSP nonce approach**: The inline `nonce` attribute on `<canvas>` won't work because `style-src` CSP doesn't apply to canvas element dimensions. The fix should set `canvas.style.height` via JS in a `useEffect` or use a CSS class instead.
2. **Height value**: The 150px default is a browser default for `<canvas>` elements without explicit dimensions. The fix should derive the height from the container or use `100%` rather than hardcoding a pixel value.

Let me know if you'd like to discuss the approach — happy to help.

0xNyk:
Thanks for identifying the CSP nonce issue, @brixyy! We've implemented the fix in #425 which supersedes this PR. The approach uses `unsafe-inline` for style-src directives (nonce must be dropped since reagraph injects `<style>` without one) and extends the fitNodesInView retries.

0xNyk:
Superseded by #425

PR #439, opened 2026-03-18
fix: session cookie SameSite + skip static assets in proxy

jadye527:
## Summary
- Change session cookie `SameSite` from `strict` to `lax` — mobile browsers behind reverse proxy tunnels (Cloudflare, ngrok, etc.) treat navigation as cross-site, blocking strict cookies after login redirect.
- Add early return in `proxy()` for static asset paths (`_next/static`, `_next/image`, `favicon.ico`, `brand/`) so they bypass auth checks and are served directly by Next.js.

## Test plan
- [ ] Login via reverse proxy tunnel on mobile browser
- [ ] Verify CSS/JS loads correctly on login page (no unstyled page)
- [ ] Verify static assets don't redirect to /login

### DIFF
--- a/src/lib/session-cookie.ts
+++ b/src/lib/session-cookie.ts
@@ -40,7 +40,7 @@ export function getMcSessionCookieOptions(input: { maxAgeSeconds: number; isSecu
   return {
     httpOnly: true,
     secure,
-    sameSite: 'strict',
+    sameSite: 'lax',
     maxAge: input.maxAgeSeconds,
     path: '/',
   }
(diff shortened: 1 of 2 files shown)


0xNyk:
Thanks for the report and patch. I’m not merging this as-is: the static-asset bypass is already handled by the proxy matcher on `main`, so that part is redundant, and the `SameSite` change needs a focused reproduction and regression coverage on its own. Closing this one for now; if you can open a focused issue/PR just for the mobile reverse-proxy cookie behavior with a reproducible test case, I’m happy to revisit it.

PR #440, opened 2026-03-18
feat: add circuit breaker for task dispatch

jadye527:
## Summary
- Prevents runaway dispatch loops when an agent repeatedly fails
- Tracks failures per task with exponential backoff
- Moves tasks to error state after threshold exceeded

## Test plan
- [ ] Assign task to offline agent, verify retries stop after threshold
- [ ] Verify successful dispatch resets failure count

### DIFF
--- a/src/lib/circuit-breaker.ts
+++ b/src/lib/circuit-breaker.ts
@@ -1,0 +1,208 @@
+/**
+ * Circuit Breaker for MC Task Dispatch
+ *
+ * Prevents retry storms that exhaust all model quotas:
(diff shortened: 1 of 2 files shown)


0xNyk:
Thanks for the contribution. I’m closing this as not mergeable in its current form because it introduces an in-memory global dispatch limiter that can block unrelated tasks after 30 dispatches/hour, resets on restart, and doesn’t include persistence or tests for the new retry semantics. This needs a design issue first so we can define the failure model and storage expectations before landing code.

PR #442, opened 2026-03-18
feat: add task dependencies with dispatch blocking

jadye527:
## Summary
- Tasks can declare dependencies on other tasks
- Dispatcher checks all dependencies are resolved before dispatching
- New API route for managing task dependencies
- DB migration for task_dependencies table

## Test plan
- [ ] Create Task A and B, make B depend on A
- [ ] Verify B stays blocked until A completes
- [ ] Complete A, verify B becomes dispatchable


0xNyk:
Thanks for the contribution. I’m closing this for now because it lands schema/API behavior without tests or UI integration, and it also claims migration id `042`, which conflicts with another open schema change. Task dependency support is a reasonable feature, but it needs a tracked issue and a fuller design/validation pass before it can merge.

PR #512, opened 2026-03-26
fix: hermes agent install fails in no-TTY environment (closes /dev/tty error)

Shipu:
## Problem

When installing Hermes Agent from the Mission Control Settings page, the install fails with:

```
main: line 1005: /dev/tty: No such device or address
```

### Root Cause

The Hermes install script checks `[ -e /dev/tty ]` to detect terminal availability. However, `/dev/tty` **always exists as a character device node** — even in non-TTY environments (Node.js child_process, CI, Docker). So `-e` always returns true even though `/dev/tty` cannot actually be opened. The script then crashes trying to read from it.

There are **3 affected locations** in the script:
- ~line 497: package install prompts
- ~line 934: setup wizard
- ~line 999: gateway setup

## Fix

Download the script first, patch all 3 occurrences of `[ -e /dev/tty ]` with a real open-test, then run the patched script:

```bash
curl -fsSL https://...install.sh -o /tmp/hermes-install.sh
sed -i 's/\[ -e \/dev\/tty \]/{ exec 3<\/dev\/tty; } 2>\/dev\/null \&\& exec 3>\&-/g' /tmp/hermes-install.sh
bash /tmp/hermes-install.sh --skip-setup
```

`{ exec 3</dev/tty; } 2>/dev/null && exec 3>&-` actually **attempts to open** `/dev/tty` — it returns false in non-TTY environments. The installer's own non-interactive guards then trigger correctly at all 3 locations.

Also: `runCommand()` in `command.ts` now always resolves (never rejects) on process exit, so `stdout`/`stderr` output is always captured regardless of exit code.

## Result

Hermes installs fully non-interactively ✅. Verified working on Ubuntu with Mission Control spawning the install as a child process.

Post-install, users can run:
- `hermes setup` — configure messaging platforms  
- `hermes gateway install` — install as a background service

## Testing

- `npm run build` ✅
- Hermes Agent installs successfully via Mission Control UI ✅

### DIFF
--- a/src/lib/agent-runtimes.ts
+++ b/src/lib/agent-runtimes.ts
@@ -396,7 +396,12 @@ async function installHermesLocal(job: InstallJob): Promise<void> {
   job.output += '> Installing Hermes Agent via official installer...\n'
   const env = getInstallEnv()
   try {
-    const result = await runCommand('bash', ['-c', 'curl -fsSL https://raw.githubusercontent.com/NousResearch/hermes-agent/main/scripts/install.sh | setsid bash -s -- --skip-setup'], {
+    const hermesInstallCmd = [
+      `curl -fsSL https://raw.githubusercontent.com/NousResearch/hermes-agent/main/scripts/install.sh -o /tmp/hermes-install.sh`,
+      `sed -i 's/\\[ -e \\/dev\\/tty \\]/{ exec 3<\\/dev\\/tty; } 2>\\/dev\\/null \\&\\& exec 3>\\&-/g' /tmp/hermes-install.sh`,
+      `bash /tmp/hermes-install.sh --skip-setup`,
+    ].join(' && ')
+    const result = await runCommand('bash', ['-c', hermesInstallCmd], {
       timeoutMs: 600_000, env,
     })
     if (result.stdout) job.output += result.stdout + '\n'


0xNyk (changes requested):
Blocking before merge:

1) `runCommand` semantics changed globally (non-zero exit now resolves instead of rejects). This is a cross-cutting behavioral change with broad risk, and can silently alter error handling in API routes that currently rely on thrown failures.
2) Installer patch path uses inline `sed -i` script rewriting of upstream install.sh. This is portability-fragile and creates maintenance risk.
3) Please split this PR:
   - A minimal Hermes no-TTY install fix, scoped to installer invocation behavior only
   - Any command-runtime semantic change in a separate PR with dedicated migration + tests for all callsites

Please resubmit with a narrowly scoped patch and explicit regression tests for no-TTY install path.

0xNyk:
Triage enforcement update from maintainers:\n\nStatus: not mergeable in current form.\n\nBlocking reasons:\n- PR is conflicting with `main`.\n- Proposed change alters global `runCommand` semantics (high blast radius).\n- No proof that no-TTY fix can be achieved without cross-cutting behavior change.\n\nRequired path forward:\n1) Rebase and isolate a local no-TTY handling fix in installer path only.\n2) Preserve existing global `runCommand` success/failure contract.\n3) Add regression tests for non-interactive environments.\n\nRecommendation: replace with a narrow fix PR.\n

0xNyk:
Tracking replacement scope in #538. Recommend closing this PR in favor of the scoped implementation issue.

0xNyk:
Closing — the `agent-runtimes.ts` fix for no-TTY environments is a good idea and we'll incorporate it. However, the `command.ts` change to always resolve (never reject on non-zero exit) is a breaking behavioral change that would silently swallow all command failures across the codebase. We'll take the install fix and implement it without modifying `runCommand` semantics. Thanks for identifying the /dev/tty issue.

PR #517, opened 2026-03-27
fix(ui): eliminate panel-navigation flicker caused by ErrorBoundary key remount

oliviercp:
## Summary

- Replaces `key={activeTab}` on `<ErrorBoundary>` in `page.tsx` with a new `resetKey` prop
- Adds `componentDidUpdate` to `ErrorBoundary` that clears error state when `resetKey` changes — without remounting children
- Every panel navigation previously destroyed and recreated the entire content subtree (React's `key` contract), producing a visible blank flash between panels

## Root cause

`<ErrorBoundary key={activeTab}>` was used to reset error state on panel change, but React's `key` mechanism achieves this by fully unmounting and remounting the wrapped subtree on every navigation — causing a blank flicker.

## Fix

`resetKey` achieves the same error-state reset via `componentDidUpdate`, leaving the existing DOM intact during normal navigation so React can reconcile only what changed.

## Test plan

- [ ] Navigate between panels — no blank flash between transitions
- [ ] Trigger a panel error, then navigate to another panel — error UI clears correctly on the new panel
- [ ] Trigger a panel error, click "Try Again" — error UI clears in place

Closes #2

### DIFF
--- a/src/app/[[...panel]]/page.tsx
+++ b/src/app/[[...panel]]/page.tsx
@@ -399,7 +399,7 @@ export default function Home() {
           aria-hidden={showOnboarding}
         >
           <div aria-live="polite" className="flex flex-col min-h-full">
-            <ErrorBoundary key={activeTab}>
+            <ErrorBoundary resetKey={activeTab}>
               <ContentRouter tab={activeTab} />
             </ErrorBoundary>
           </div>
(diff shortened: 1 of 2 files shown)


0xNyk:
Superseded by #519; closing duplicate thread.

PR #524, opened 2026-03-29
fix(dispatch): persistent session tasks skip Aegis bounce loop

jadye527:
## Summary
- When a task has `target_session` set, `chat.send` is fire-and-forget. Previously the confirmation string became the `resolution` and Aegis rejected it, causing an infinite `assigned→in_progress→review` bounce loop.
- Fix: after successful `chat.send`, write `persistent_session=true` to metadata, insert scheduler comment, `recordSuccess`, and `continue` — never fall through to the Aegis review path
- Update `buildTaskPrompt()` to add `RESOLUTION:` signal instructions when `persistentSession=true`
- Move `taskMeta` read before `buildTaskPrompt` so the persistent flag can be passed in

## Test plan
- [ ] Deploy to dev (port 3003) and dispatch a task with `target_session` set
- [ ] Verify task stays `in_progress` after dispatch (not bouncing to `review`)
- [ ] Verify scheduler comment inserted: "Task dispatched to session X. Waiting for RESOLUTION: signal"
- [ ] Verify metadata has `persistent_session: true`

## Files changed
- `src/lib/task-dispatch.ts` — persistent session early-exit + prompt update

### DIFF
--- a/src/lib/task-dispatch.ts
+++ b/src/lib/task-dispatch.ts
@@ -37,7 +37,7 @@ function resolveGatewayAgentId(task: DispatchableTask): string {
   return task.agent_name
 }
 
-function buildTaskPrompt(task: DispatchableTask, rejectionFeedback?: string | null): string {
+function buildTaskPrompt(task: DispatchableTask, rejectionFeedback?: string | null, persistentSession?: boolean): string {
   const ticket = task.ticket_prefix && task.project_ticket_no
     ? `${task.ticket_prefix}-${String(task.project_ticket_no).padStart(3, '0')}`
     : `TASK-${task.id}`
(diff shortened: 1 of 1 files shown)


0xNyk:
Closing this PR as non-mergeable in current form.

Reason:
- Diff is heavily contaminated with unrelated/generated/runtime artifacts (`.cloudflared/*`, `*.pid`, logs, `mission-control.db`, `dist-scripts/*`, etc.) and too large to safely review as a bugfix.

Please open a fresh, minimal PR that includes only the persistent-session dispatch fix and associated tests.

0xNyk:
Closed pending clean re-PR with minimal scoped diff.

PR #462, opened 2026-03-21
fix: security hardening from audit

0xNyk:
## Summary

Defense-in-depth security improvements discovered during audit prompted by #460. No malware or credential leaks were found, but the audit uncovered real hardening opportunities.

### Critical/High
- **Session token hashing**: Store SHA-256 hash of session tokens in DB instead of plaintext. DB compromise no longer yields session hijack. Migration `043` hashes existing tokens (invalidates active sessions — users re-login once).
- **Password minimum length**: Self-service password change now enforces 12 chars (was 8; creation already required 12).
- **Proxy auth IP enforcement**: New `MC_PROXY_AUTH_TRUSTED_IPS` env var restricts which IPs can send the proxy auth header, preventing header spoofing from direct clients.

### Medium
- **Scrypt cost increase**: N=16384→65536 (2^14→2^16) with progressive rehash on login — existing passwords upgrade transparently.
- **HSTS default**: Enabled by default in production. Opt-out via `MC_DISABLE_HSTS=1` (was opt-in via `MC_ENABLE_HSTS=1`).
- **Debug SSRF restriction**: POST proxy endpoint restricted to allowlisted gateway API paths.
- **Session cookie secure flag**: Defaults to `true` in production.
- **Rate limit disable guard**: `MC_DISABLE_RATE_LIMIT` only works when `NODE_ENV !== 'production'`.
- **Log sanitization**: Removed interpolated password value from insecure-default warning.
- **Docker secrets permissions**: `chmod 600` on generated secrets file.

## Test plan
- [x] All 870 unit tests pass
- [x] TypeScript type check passes
- [x] ESLint: 0 errors
- [x] Build succeeds
- [ ] Manual: login/logout flow works after deployment
- [ ] Manual: password change enforces 12 chars
- [ ] Manual: existing session invalidated after migration (re-login required)

PR #568, opened 2026-04-08
fix: prevent X-Forwarded-For IP spoofing in proxy auth

AlessandroMartini:
## Summary

- **Fix X-Forwarded-For IP spoofing** in proxy auth (`getUserFromRequest`). The current code takes the **first (leftmost)** IP from `X-Forwarded-For`, which is attacker-controlled. An adversary can prepend a trusted proxy IP to bypass the `MC_PROXY_AUTH_TRUSTED_IPS` gate and impersonate any user, including admin.
- Walk the `X-Forwarded-For` chain **right-to-left**, skipping known trusted proxy IPs, to find the actual client IP — consistent with the approach already used in `extractClientIp()` (`src/lib/rate-limit.ts`).

## Security impact

Without this fix, when `MC_PROXY_AUTH_TRUSTED_IPS` is configured, any client can send:
```
X-Forwarded-For: <trusted-proxy-ip>, <real-attacker-ip>
```
and the code picks `<trusted-proxy-ip>` as `clientIp`, passing the `trustedIps.has(clientIp)` check. The attacker then controls the proxy auth header and assumes any identity.

## Regression introduced in

PR #462 (`fix: security hardening from audit`) added `MC_PROXY_AUTH_TRUSTED_IPS` but used `split(',')[0]` (leftmost) instead of walking right-to-left.

## Test plan

- [x] `pnpm typecheck` passes
- [ ] Verify with `X-Forwarded-For: <trusted>, <attacker>` — should reject
- [ ] Verify with `X-Forwarded-For: <attacker>, <trusted>` — should reject (attacker is rightmost untrusted)
- [ ] Verify legitimate proxy chain still works

### DIFF
--- a/src/lib/auth.ts
+++ b/src/lib/auth.ts
@@ -424,9 +424,19 @@ export function getUserFromRequest(request: Request): User | null {
   if (proxyAuthHeader) {
     const trustedIps = PROXY_AUTH_TRUSTED_IPS
     if (trustedIps.size > 0) {
-      const clientIp = request.headers.get('x-real-ip')?.trim()
-        || request.headers.get('x-forwarded-for')?.split(',')[0]?.trim()
-        || ''
+      // Walk X-Forwarded-For right-to-left, skipping trusted proxy IPs,
+      // to find the actual client IP (prevents spoofing via prepended headers).
+      const xff = request.headers.get('x-forwarded-for')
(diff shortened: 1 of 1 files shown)


AlessandroMartini:
Superseded by consolidated PR — see below.

PR #592, opened 2026-04-16
fix: resolve CSRF origin mismatch behind reverse proxies

mxihan:
## Summary                                                                
  - CSRF validation now checks all host candidates (x-forwarded-host,       
  x-original-host, forwarded, etc.) instead of only the Host header         
  - Adds port-normalized comparison that strips default ports (80/443)      
  before matching                                                           
                                                 
  ## Why                                                                    
  When running behind Nginx Proxy Manager with a custom HTTPS port (e.g. 
  8443), the browser sends `Origin: https://host:8443` but the proxy may    
  forward `Host: host` (without port). This caused false-positive CSRF 403 
  errors on all mutating requests (login, etc.).

### DIFF
--- a/src/proxy.ts
+++ b/src/proxy.ts
@@ -84,6 +84,29 @@ function hostMatches(pattern: string, hostname: string): boolean {
   return h === p
 }
 
+/** Normalize a host:port string by stripping default ports (80 for http, 443 for https). */
+function stripDefaultPort(host: string): string {
+  const h = host.toLowerCase()
+  if (h.endsWith(':443')) return h.slice(0, -4)
+  if (h.endsWith(':80')) return h.slice(0, -3)
+  return h
+}
+
+/**
+ * Compare a request host candidate with the Origin host for CSRF validation.
+ * Handles port mismatches caused by reverse proxies (e.g. Origin includes :8443
+ * but the Host header may have been rewritten or stripped by the proxy).
+ */
+function hostsMatchForCsrf(requestHost: string, originHost: string): boolean {
+  const a = normalizeHostname(requestHost)
+  const b = normalizeHostname(originHost)
+  if (!a || !b) return false
+  // Exact match first
+  if (a === b) return true
+  // Match after stripping default ports
+  return stripDefaultPort(a) === stripDefaultPort(b)
+}
+
 function nextResponseWithNonce(request: NextRequest): { response: NextResponse; nonce: string } {
   const nonce = crypto.randomBytes(16).toString('base64')
   const googleEnabled = !!(process.env.NEXT_PUBLIC_GOOGLE_CLIENT_ID || process.env.GOOGLE_CLIENT_ID)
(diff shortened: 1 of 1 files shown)


0xNyk (changes requested):
## Security Audit: REQUEST CHANGES

### Critical: CSRF bypass via proxy header spoofing

The PR replaces the single `Host` header comparison with comparison against **ALL** proxy headers from `getRequestHostCandidates()` (X-Forwarded-Host, X-Original-Host, etc.). A malicious client can set `X-Forwarded-Host: attacker.com` on their request. If their `Origin` header is also `https://attacker.com`, the CSRF check passes — completely bypassing protection.

**Recommended fix**: For CSRF validation, only compare Origin against the `Host` header (as before), but apply `normalizeHostname` to both sides to fix the port mismatch. Do NOT trust proxy headers for CSRF unless explicitly configured via an opt-in setting like `MC_TRUST_PROXY_HEADERS`.

### `stripDefaultPort` is dead code

`normalizeHostname()` already strips the port via `.split(":")[0]`. After normalization there are never any ports left, so `stripDefaultPort()` on the result is a no-op. The actual fix works purely because of `normalizeHostname`, not `stripDefaultPort`. Remove it.

### Missing tests

No unit tests were added for `hostsMatchForCsrf` or the CSRF check flow.

### Pre-existing: IPv6 broken

`normalizeHostname("[::1]:3000")` returns empty string. This should be fixed since the PR touches this code path.

pachecocordovamoiseseduardo-byte (approved):

0xNyk (approved):
Approving — and walking through the security check since this PR touches the CSRF surface.

**Concern I started with:** "reverse proxy origin handling" is a classic place to introduce a CSRF bypass disguised as a fix — if the new code trusts `X-Forwarded-Host` blindly, the fix IS the vulnerability.

**Walked through the code path:**

1. `getRequestHostCandidates()` (existing in `src/proxy.ts`) builds candidates from `X-Forwarded-Host`, `X-Original-Host`, `X-Forwarded-Server`, the `Forwarded` header, `Host`, and `request.nextUrl.host`.
2. Those candidates are first run through the **host allowlist** (`MC_ALLOWED_HOSTS` in production, default-deny via `enforceAllowlist`). Any request whose candidates aren't in the allowlist is rejected with `Forbidden` before CSRF validation runs at all.
3. Only AFTER passing the allowlist does the new CSRF check (`hostsMatchForCsrf`) compare Origin against the same candidate list with port-stripping for default ports.

So the existing two-layer architecture (allowlist → CSRF) means `X-Forwarded-Host` can only contribute a candidate for the CSRF match if that host is already in `MC_ALLOWED_HOSTS`. No bypass.

**One thing operators should know** (already documented in the existing code path, but worth restating): in production, `MC_ALLOWED_HOSTS` MUST be set, otherwise the CSRF check effectively trusts whatever Origin matches the request `Host` header. That's the existing contract; this PR doesn't change it.

Merging.

PR #601, opened 2026-04-20
fix(agents): flatten gateway_config into buildAgentConfig overrides

### LINKED ISSUE #599: [Bug] Creating an agent from a template double-wraps `config.model.primary`, crashing the Models/Config tabs
## Describe the Bug

Creating an agent from a template via the UI writes a malformed `config.model` to SQLite: `config.model.primary` is stored as an object `{ primary: "<model-id>" }` instead of a plain string. When the newly-created agent is opened, the **Models** and **Config** tabs crash the React tree with:

> Objects are not valid as a React child (found: object with keys `{primary}`). If you meant to render a collection of children, use an array instead.

Root cause (code pointers):

- The UI submits the documented payload shape — `gateway_config: { model: { primary: <string> }, identity: { ... }, sandbox: { ... } }` — from `src/components/panels/agent-detail-tabs.tsx` (create flow).
- The API route then passes this nested object straight into `buildAgentConfig`, whose overrides contract is **flat** (`model: string`). The type mismatch is silenced with `as any`:

  ```ts
  // src/app/api/agents/route.ts:193
  const builtConfig = buildAgentConfig(tpl, (gateway_config || {}) as any);
  ```

  ```ts
  // src/lib/agent-templates.ts:447
  if (overrides.model) config.model.primary = overrides.model
  ```

  Because `overrides.model` is actually `{ primary: "anthropic/..." }` (not a string), the whole object gets assigned to `config.model.primary`. Several other nested fields (`identity.*`, `sandbox.*`) also silently fall through for the same reason; only template defaults make it in.

- `ModelsTab` at `src/components/panels/agent-detail-tabs.tsx:2876` and the structured view of `ConfigTab` at `~:1701` render `primary` directly as a React child, so any agent created this way crashes both tabs on open.

Suggested fix: flatten `gateway_config` into the shape `buildAgentConfig` actually expects, and drop the `as any` (per `CONTRIBUTING.md` style rule). Add a unit test asserting `config.model.primary` remains a string when `overrides.model` is provided. Happy to send a PR.

## Steps to Reproduce

1. Fresh clone `main`, `pnpm install`, `pnpm dev`.
2. Complete first-run setup at `/setup` to create an admin.
3. In the UI, create a new agent from any built-in template (e.g. developer / sonnet tier), keeping the default model.
4. Open the newly-created agent's detail panel.
5. Click the **Models** tab (or **Config** tab).

## Expected Behavior

Models/Config tab renders normally, showing the primary model as a plain string (e.g. `anthropic/claude-sonnet-4-20250514`) and the fallback list. The stored row in `.data/mission-control.db` should contain `config.model.primary` as a string.

## Actual Behavior

The React error boundary fires with: *"Objects are not valid as a React child (found: object with keys `{primary}`)."*

Inspecting the row directly confirms the double-wrap:

```json
"model": {
  "primary": { "primary": "anthropic/claude-sonnet-4-20250514" },
  "fallbacks": [ "openrouter/anthropic/claude-sonnet-4", "..." ]
}
```

## Screenshots

Image

## Environment

- OS: Windows 11 (10.0.26200)
- Browser: Chrome Version 147.0.7727.56 (Official Build) (64-bit)
- Mission Control Version: 2.0.1 (`main` @ commit `503a289`)
- Node: 24.12.0
- pnpm: 9.12.1
- Next.js: 16.1.6 (Turbopack)
- better-sqlite3: 12.6.2

codingDr:
# Summary
Fixes #599 — crash in the Models and Config tabs for any agent created from a built-in template.

Root cause: `src/app/api/agents/route.ts` passed the nested `gateway_config` object straight into `buildAgentConfig` with an `as any` cast. The function declares a flat overrides contract (`model: string`, etc.), so the cast hid a real mismatch: `overrides.model` was `{ primary: "<id>" }`, which then got assigned wholesale to `config.model.primary` — leaving the DB with a doubly-wrapped primary. `ModelsTab` and `ConfigTab` render `primary` as a React child, so both tabs throw *"Objects are not valid as a React child (found: object with keys {primary})"* on open.

Fix: extract the leaf fields from `gateway_config` at the call site so the primary string is preserved and the other overrides (identity, sandbox, subagents) actually take effect.

See #599 for full repro and DB evidence.

# Risk Level
Low — scoped to one call site in the agent-create route. Behavior change is strictly "`model.primary` is now a string as documented." Pre-existing agents with the malformed shape need a one-time row fix (noted below).

# Tests
- `pnpm test src/lib/__tests__/agent-templates.test.ts` — new unit test, 5 passed
- `pnpm lint src/app/api/agents/route.ts src/lib/__tests__/agent-templates.test.ts` — clean
- `pnpm typecheck` — clean
- Full `pnpm test` — only pre-existing, unrelated Windows path / OpenCode failures

# Contribution Checklist
- [x] Tests added/updated for behavior changes
- [x] Lint/typecheck passing (build/e2e: please run in CI)
- [x] Security review done if auth/data/crypto touched (N/A)
- [x] DB migration tested if schema changed (no schema change)

# Notes
Existing rows written with the bug still crash the UI. Operators can fix with a one-off SQL update, e.g.:

```sql
UPDATE agents
   SET config = json_set(config, '$.model.primary',
                         json_extract(config, '$.model.primary.primary'))
 WHERE json_valid(config)
   AND json_type(config, '$.model.primary') = 'object';
```

# Screenshots
Tabs tested after applying fix:

models-tab

config-tab

### DIFF
--- a/src/app/api/agents/route.ts
+++ b/src/app/api/agents/route.ts
@@ -190,7 +190,21 @@ export async function POST(request: NextRequest) {
     if (template) {
       const tpl = getTemplate(template);
       if (tpl) {
-        const builtConfig = buildAgentConfig(tpl, (gateway_config || {}) as any);
+        // gateway_config uses the nested OpenClaw shape; flatten to buildAgentConfig's
+        // overrides contract. Passing the nested object directly (as any) caused
+        // config.model.primary to be set to { primary: "..." } instead of the string.
+        const gc = (gateway_config || {}) as any;
+        const builtConfig = buildAgentConfig(tpl, {
+          id: openclawId,
+          name,
+          model: gc.model?.primary,
+          emoji: gc.identity?.emoji,
+          theme: gc.identity?.theme,
+          workspaceAccess: gc.sandbox?.workspaceAccess,
+          sandboxMode: gc.sandbox?.mode,
+          dockerNetwork: gc.sandbox?.docker?.network,
+          subagentAllowAgents: gc.subagents?.allowAgents,
+        });
         finalConfig = { ...builtConfig, ...finalConfig };
         if (!finalRole) finalRole = tpl.config.identity?.theme || tpl.type;
       }
--- a/src/lib/__tests__/agent-templates.test.ts
+++ b/src/lib/__tests__/agent-templates.test.ts
@@ -1,0 +1,45 @@
+import { describe, expect, it } from 'vitest'
+import { AGENT_TEMPLATES, buildAgentConfig, getTemplate } from '@/lib/agent-templates'
+
+describe('buildAgentConfig', () => {
+  const tpl = AGENT_TEMPLATES[0]
+
+  it('sets config.model.primary as a string when overrides.model is a string', () => {
+    const result = buildAgentConfig(tpl, {
+      id: 'test',
+      name: 'Test',
+      model: 'anthropic/claude-sonnet-4-20250514',
+    })
+    expect(typeof result.model.primary).toBe('string')
+    expect(result.model.primary).toBe('anthropic/claude-sonnet-4-20250514')
+  })
+
+  it('preserves the template default model when no override is provided', () => {
+    const result = buildAgentConfig(tpl, { id: 'test', name: 'Test' })
+    expect(typeof result.model.primary).toBe('string')
+    expect(result.model.primary).toBe(tpl.config.model.primary)
+  })
+
+  it('applies identity overrides without wrapping them', () => {
+    const result = buildAgentConfig(tpl, {
+      id: 'test',
+      name: 'Test',
+      emoji: 'X',
+      theme: 'some theme',
+    })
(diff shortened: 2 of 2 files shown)


0xbrainkid (approved):
Approved: flattens gateway_config into buildAgentConfig's expected override shape and adds regression tests for model/identity overrides.

0xNyk (approved):
Approving — fixes #599. Root cause: passing the nested `gateway_config` straight into `buildAgentConfig` with `as any` cast meant `config.model.primary` ended up as `{ primary: "..." }` (a wrapped object) instead of a string, crashing the React tree on the Models/Config tabs. Fix flattens the nested shape into the function's flat-overrides contract.

Tests verify that `config.model.primary` is a string after build, that template defaults survive when no override is given, and that identity overrides apply without wrapping. Closes #599.

PR #605, opened 2026-04-21
fix: disable caching on release check endpoint and client fetch

dizhaky:
## Summary
- disable caching on the release check endpoint
- disable cached client fetch behavior for the same path

## Notes
- opened from the synced fork branch after rebasing onto upstream `main`
- draft PR for upstream review

### DIFF
--- a/src/app/[[...panel]]/page.tsx
+++ b/src/app/[[...panel]]/page.tsx
@@ -232,7 +232,7 @@ export default function Home() {
       .catch(() => { markStep('auth') })
 
     // Check for available updates
-    fetch('/api/releases/check')
+    fetch('/api/releases/check', { cache: 'no-store' })
       .then(res => res.ok ? res.json() : null)
       .then(data => {
         if (data?.updateAvailable) {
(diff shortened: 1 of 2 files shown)


0xbrainkid:
NOTE: draft PR reviewed only as a note, not an approval. The no-store change is directionally fine for avoiding stale release checks, but please mark ready-for-review after confirming whether this endpoint should use no-store on both success and failure responses or a very short private cache to avoid unnecessary GitHub API calls.

liberathio (approved):
Tiny + low-risk change. Reviewed: replaces 1h cache on /api/releases/check with no-store. The catch block already handles upstream GitHub rate limits gracefully, so no functional regression. LGTM.

liberathio:
## Review (community / read-only) — LGTM ✅

Tiny + low-risk. Replaces `Cache-Control: public, max-age=3600` on `/api/releases/check` with `no-store` (also on the client `fetch`).

**Trade-off:**
- **Pro:** users see new releases immediately after publish.
- **Con:** more calls to the GitHub releases API per session. The endpoint's existing `catch` already returns `updateAvailable: false` gracefully on rate-limit failures, so the worst-case is "no update notification temporarily" — not a hard error.

**Verdict:** Safe to merge for single-user / self-hosted setups. Multi-tenant deployments behind a shared egress IP might want to keep some short cache (e.g. `max-age=60`) to stay under GitHub's unauthenticated 60 req/h, but that's an optimization, not a correctness issue.

(Reviewer is read-only on this repo — approval doesn't count toward `REVIEW_REQUIRED`; flagging for a maintainer.)

PR #607, opened 2026-04-22
mission-control:v2.0.1-v0.9 - Replaced both callGatewayAgent(...) cal…

jmmc-tools:
# Summary

Closes #608 
Replace the CLI-subprocess-based openclaw gateway integration with a native
WebSocket client and an HTTP chat completions transport for task/aegis dispatch.

Previously `callOpenClawGateway` shelled out to the `openclaw` CLI binary and
`callGatewayAgent` used `agent` + `agent.wait` RPC which only returns lifecycle
metadata — the LLM response text was never surfaced to MC. Tasks would
accumulate `dispatch_attempts` and never transition out of `assigned`.

Changes:
- **`callOpenClawGateway`**: now a native `ws` WebSocket client — no CLI dependency,
  correct auth handshake, `translateMethod()` maps all legacy method names to
  gateway RPC dot-notation transparently.
- **`callGatewayAgentHTTP`** (new): uses `POST /v1/chat/completions` with model
  `openclaw/<agentId>` — returns full text via `choices[0].message.content`.
- **`task-dispatch.ts`**: both task dispatch and aegis review switched from
  `runOpenClaw([…gateway call agent…])` to `callGatewayAgentHTTP(…, 300_000)`.
- **`Dockerfile`**: 3-stage build — openclaw binary copied from its own image
  (stage 0), not downloaded at runtime.

Verified in production: both test tasks transitioned to `status=done, outcome=success`
with full LLM text stored as comments.

# Risk Level
Medium

The gateway communication layer is completely replaced. The new WS client is
protocol-compatible (tested against the same gateway). HTTP completions endpoint
was confirmed working before being adopted as the dispatch path. The `parseGatewayJsonOutput`
export is kept for test backward compatibility.

# Tests

```bash
# Build and push
docker build --platform linux/amd64 \
  -t builderz-labs/mission-control:v2.0.1-v0.9 \
  /path/to/mission-control
docker push builderz-labs/mission-control:v2.0.1-v0.9

# Deploy
kubectl apply -f dep_all_in_one.yaml
kubectl rollout status deployment/mission-control -n mission-control
# → successfully rolled out

# Reset tasks and observe dispatch
kubectl exec -n mission-control <pod> -- node -e "
  const db = require('better-sqlite3')('/app/.data/mission-control.db');
  db.prepare(\"UPDATE tasks SET status='assigned', dispatch_attempts=0 WHERE id IN (1,2)\").run();
"

# Verify result (60–90 seconds later)
[...]

PR #610, opened 2026-04-23
[Bug] Fixed chat sessions via openclaw gateway

jmmc-tools:
# Summary
related #611 
**feat(chat): agent chat via openclaw HTTP completions + agents sidebar**

Two changes to make the Mission Control chat panel actually work with openclaw agents:

1. **Agents sidebar** (`conversation-list.tsx`): Added an "Agents" section above Sessions
   that lists all registered agents with status dot and role. Clicking one sets
   `activeConversation = "agent_<name>"` and enables the input — no WebSocket needed.
   Fixed the `_onNewConversation` alias that was silently ignoring the prop.

2. **HTTP completions fallback** (`/api/chat/messages` POST): When a message is forwarded
   to an agent that has no active gateway session, MC now calls the openclaw gateway's
   `/v1/chat/completions` endpoint (`model: "openclaw/<agentId>"` → fallback `"openclaw"`)
   instead of fire-and-forget CLI. The response is persisted via `createChatReply` and
   broadcast over SSE so the browser renders it immediately.
   Also removed the silent `no_active_session` dead-end — every agent message now gets
   a response or a visible error status.

Root cause: the existing code used `runOpenClaw` CLI for the no-session path but never
captured the response, so messages were stored but replies never appeared.

# Risk Level

Low — additive changes only. Existing session-based (`sessionKey`) and coordinator flows
are untouched. The HTTP completions call is a new else-branch with its own error handling.

# Tests

```bash
# Verify completions endpoint responds
kubectl exec -n openclaw-poc <pod> -- curl -s \
  -H 'Authorization: Bearer <token>' \
  -H 'Content-Type: application/json' \
  -d '{"model":"openclaw","messages":[{"role":"user","content":"hola"}],"stream":false}' \
  http://localhost:18789/v1/chat/completions
# → {"choices":[{"message":{"content":"¡Hola!"}}]}


```

Typecheck: `pnpm tsc --noEmit` — no errors.

# Contribution Checklist

- [ ] Security review done if auth/data/crypto touched

# Notes

- The openclaw gateway must have `http.endpoints.chatCompletions.enabled: true` in its
  config (already set in the deployment ConfigMap).
- `OPENCLAW_GATEWAY_HOST` / `OPENCLAW_GATEWAY_PORT` / `OPENCLAW_GATEWAY_TOKEN` env vars
  must be set in the MC pod — these are read from `src/lib/config.ts` and
  `src/lib/gateway-runtime.ts` for the HTTP call.
- Conversation history (last 20 messages) is included in each completions request for
  context. Adjust the `LIMIT 20` in the query if needed.
- The `openclawId: "main"` in the test2 agent config causes the first attempt to use
  `model: "openclaw/main"` — it falls back to `model: "openclaw"` automatically on error.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "docker.commands.build": "docker build --platform linux/amd64 ${dockerfile} ${tag} ${context}"
+}
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1.7
-# ── Stage 0: extraer binario openclaw desde su imagen ──────────
-FROM logstreamingacrpro.azurecr.io/openclaw:2026.4.14-secure-v0.1 AS openclaw-bin
+# ── Stage 0: extraer binario openclaw desde su imagen oficial ──
+FROM ghcr.io/openclaw/openclaw:latest AS openclaw-bin
 
 # ── Stage 1: deps ──────────────────────────────────────────────
 FROM node:22.22.0-slim AS base
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,56 +1,77 @@
+# syntax=docker/dockerfile:1.7
+# ── Stage 0: extraer binario openclaw desde su imagen ──────────
+FROM logstreamingacrpro.azurecr.io/openclaw:2026.4.14-secure-v0.1 AS openclaw-bin
+
+# ── Stage 1: deps ──────────────────────────────────────────────
 FROM node:22.22.0-slim AS base
 RUN corepack enable && corepack prepare pnpm@latest --activate
 WORKDIR /app
 
 FROM base AS deps
-# Copy only dependency manifests first for better layer caching
-COPY package.json ./
-COPY pnpm-lock.yaml* ./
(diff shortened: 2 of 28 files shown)


0xbrainkid:
REWORK: this stacks the large risky Docker/gateway rewrite from #607 with additional chat/session/workspace changes. In /api/chat/messages I also see dead placeholder control flow using if false / else, and a synchronous completions fallback that turns no-active-session chat into a 22s HTTP request without tests for timeout/error/history mapping. Next step: base this on a clean branch after the transport changes are isolated, remove the dead branch, and add tests for no-active-session, gateway timeout, model fallback, and reply persistence/SSE broadcast.

0xNyk:
Thanks — closing this one in favor of #655 + #648 which just landed.

Same reasoning as #607: those two PRs cover the gateway/dispatch fix, and the chat-related part of your work (Agents sidebar, agent\_<name> conversation routing) is worth landing separately as a focused chat-only PR — without the `fix-git-permissions.sh` and `.vscode/settings.json` scope creep that currently rides along.

If you'd like to resubmit just the conversation-list / chat-workspace changes against the latest `main`, happy to review.

Closing.

PR #621, opened 2026-04-27
fix(validation): drop default values from updateTaskSchema

galongin:
## Summary

`updateTaskSchema = createTaskSchema.partial()` makes every field optional, but Zod's `.partial()` keeps the underlying `.default(...)` modifiers. So a `PUT /api/tasks/{id}` body that omits a field parses with the create-time default applied (`status: 'inbox'`, `priority: 'medium'`, `tags: []`, `metadata: {}`).

Combined with the route's `if (field !== undefined) UPDATE …` guards in `src/app/api/tasks/[id]/route.ts`, this means a partial update like `PUT { title }` silently overwrites the stored row: status is reset to `inbox`, tags/metadata are wiped.

## Repro (before this fix)

```sh
# create a task in done state
curl -s -X POST -H "x-api-key: $K" -H 'content-type: application/json' \
  http://localhost:3000/api/tasks \
  -d '{"title":"t1","status":"done","tags":["a","b"],"metadata":{"foo":"bar"}}'
# → status=done, tags=[a,b], metadata={foo:bar}

# update only the description
curl -s -X PUT  -H "x-api-key: $K" -H 'content-type: application/json' \
  http://localhost:3000/api/tasks/<id> \
  -d '{"description":"updated"}'

# read back
curl -s -H "x-api-key: $K" http://localhost:3000/api/tasks/<id>
# → status=inbox  ❌      tags=[]  ❌     metadata={}  ❌
```

The activity log on each affected task records the silent rewrite, e.g. `Task updated: status: done → inbox, …`.

I hit this while bulk-updating tasks via the API — every PUT that didn't include `status` reset 60+ tasks to `inbox` and emptied their `tags`/`metadata`.

## Fix

Define `updateTaskSchema` explicitly with `.optional()` and no defaults. To keep the two schemas from drifting, extract the field validators into a shared `taskFields` map; defaults remain on `createTaskSchema` only.

The existing route handler is already shaped correctly (`if (field !== undefined) push(...)`) — once the validator stops injecting defaults, partial updates round-trip as expected.

No API contract change for clients that already send full bodies. Clients that previously got an unintended reset will now correctly preserve the omitted fields.

## Test plan

- [x] `pnpm test src/lib/__tests__/validation.test.ts` — added 4 regression tests:
  - empty body parses to `{}` (no defaults)
  - single-field update doesn't inject other fields
  - provided fields pass through verbatim
  - field constraints still enforced (`status: 'invalid'`, `feedback_rating: 6`, empty title)
- [x] `pnpm typecheck`
- [x] `pnpm lint` (only pre-existing warnings)
- [x] `pnpm test` — all task-related suites green; the only failure (`gateway-url.test.ts`) is pre-existing on `main` and unrelated.

## Notes

Only one `.partial()` site in `src/lib/validation.ts`; nothing else relies on `updateTaskSchema` injecting defaults. `createTaskSchema` behavior is unchanged.

### DIFF
--- a/src/lib/__tests__/validation.test.ts
+++ b/src/lib/__tests__/validation.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect } from 'vitest'
 import {
   createTaskSchema,
+  updateTaskSchema,
   createAgentSchema,
   createWebhookSchema,
   createAlertSchema,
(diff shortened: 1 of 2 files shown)


0xbrainkid (approved):
Approved: fixes the update schema regression by removing create-time defaults from partial updates, with focused validation tests covering omitted fields and constraints.

0xNyk (approved):
Approving — `updateTaskSchema = createTaskSchema.partial()` was carrying the create-time defaults (`status: 'inbox'`, `priority: 'medium'`, `tags: []`), so a PUT that omitted those fields would silently reset them. Refactor extracts the shared `taskFields` validators into a constant, applies defaults only on `createTaskSchema`, and constructs `updateTaskSchema` directly without inheriting them.

Tests cover empty PUT, partial PUT, full PUT, and validation-still-applies. Clean fix.

PR #623, opened 2026-04-27
Stabilize gateway websocket autodetect

agoticobran:
## Summary
- Add a shared gateway websocket URL/token resolver.
- Avoid leaking loopback gateway hosts when Cobran-OS is accessed from a non-loopback hostname.
- Default HTTPS origins to wss/443 unless an explicit gateway port is configured.
- Respect NEXT_PUBLIC_GATEWAY_TOKEN in the websocket handshake fallback.
- Add focused unit coverage for local, hosted, explicit URL, HTTPS, and token cases.

## Validation
- pnpm lint ✅
- pnpm typecheck ✅
- pnpm test ✅ 76 tests
- pnpm build ✅


0xbrainkid (approved):
Approved: gateway URL resolution now avoids leaking loopback env hosts to remote browsers, preserves explicit gateway URLs/tokens, and includes targeted tests.

PR #629, opened 2026-04-28
fix(gateway-url): downgrade http(s)://localhost inputs to ws:// for the gateway WebSocket

imtylervo:
## Problem

`buildGatewayWebSocketUrl` preserved any user-set protocol on localhost inputs — including `http://` and `https://`, which are not valid WebSocket schemes. Passing `host='https://127.0.0.1:18789'` returned `https://127.0.0.1:18789` verbatim instead of the expected `ws://127.0.0.1:18789`.

## Why now

Additive tightening on top of @0xnykcd's a020d1b — that change already routes the `NEXT_PUBLIC_GATEWAY_REVERSE_PROXY=1` path to `wss://` when the browser is on HTTPS, and that behavior is preserved here. The only new constraint is: for localhost, `http://` and `https://` (and bare `ws://`) collapse to `ws://`; `wss://` stays `wss://` as the operator's explicit TLS-terminator opt-in.

## Changes

- `src/lib/gateway-url.ts`: tighten the prefixed-URL localhost branch (line 97-118) — collapse non-`wss` protocols to `ws:`, leave `wss:` untouched. Bare-host path (line 124-125) and no-host path (line 85) continue to honor `NEXT_PUBLIC_GATEWAY_REVERSE_PROXY`.
- `src/lib/__tests__/gateway-url.test.ts`: 2 new cases pin the `http://localhost → ws://` and `wss://127.0.0.1 (preserved)` invariants.

Diff: +24 / −4 across 2 files.

## Tests

- New: \`downgrades http://localhost to ws://\` (http://localhost:18789 → ws://localhost:18789)
- New: \`preserves explicit wss:// on localhost (reverse-proxy TLS opt-in)\` (wss://127.0.0.1:18789 unchanged)
- Pre-existing: \`downgrades https://127.0.0.1 to ws:// when no reverse proxy is configured\` continues to assert the https://localhost path.

## Validation

- \`npx vitest run src/lib/__tests__/gateway-url.test.ts\` → 17/17 ✅
- \`npx vitest run\` → 82 files / 933 tests ✅
- \`npx tsc --noEmit\` → clean ✅

## Related

Aware of #623 (broader gateway WebSocket autodetect refactor). This PR is intentionally narrower — a one-function tightening — and is additive on top of a020d1b. If #623 lands first I'm happy to rebase or close.

### DIFF
--- a/src/lib/__tests__/gateway-url.test.ts
+++ b/src/lib/__tests__/gateway-url.test.ts
@@ -42,6 +42,22 @@ describe('buildGatewayWebSocketUrl', () => {
     })).toBe('ws://127.0.0.1:18789')
   })
 
+  it('downgrades http://localhost to ws://', () => {
+    expect(buildGatewayWebSocketUrl({
+      host: 'http://localhost:18789',
+      port: 18789,
+      browserProtocol: 'http:',
+    })).toBe('ws://localhost:18789')
+  })
+
+  it('preserves explicit wss:// on localhost (reverse-proxy TLS opt-in)', () => {
+    expect(buildGatewayWebSocketUrl({
+      host: 'wss://127.0.0.1:18789',
+      port: 18789,
+      browserProtocol: 'https:',
+    })).toBe('wss://127.0.0.1:18789')
+  })
+
   it('omits 18789 for remote hosts on https browser context', () => {
     expect(buildGatewayWebSocketUrl({
       host: 'node-01.tailnet123.ts.net',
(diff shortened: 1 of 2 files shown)


0xbrainkid (approved):
Approved: focused localhost gateway URL fix with regression coverage; http(s) localhost now resolves to ws:// while explicit wss:// remains preserved.

0xNyk (approved):
Approving — clean fix for the WebSocket URL builder. `http://localhost` and `https://localhost` now correctly downgrade to `ws://` (the gateway doesn't speak TLS locally), while `wss://` is preserved as the operator's explicit opt-in for reverse-proxy TLS termination. Tests cover both the http→ws downgrade and the wss preservation. Two-file diff with no surprises.

PR #631, opened 2026-04-29
fix: send user id in DELETE request body

### LINKED ISSUE #614: [Bug] Delete user from UI fails with "Request body required"
## Description

Attempting to delete a user from the MC dashboard (Settings → Users) fails with the error: `"Request body required"`. The API endpoint `DELETE /api/auth/users` works correctly when called with a JSON body containing the user ID, but the UI does not send the required body.

## Steps to Reproduce

1. Log in as admin
2. Go to Settings → Users (or Session → Users)
3. Attempt to delete a non-self user
4. Error: `"Request body required"`

## Expected Behavior

The delete button in the UI should send the correct request body with the user ID.

## API Details

The backend expects the user ID in the request body:
```typescript
// /src/app/api/auth/users/route.ts
const body = await request.json()
const { id } = body
```

But the frontend likely sends a DELETE request without a body or with the ID in the URL instead.

## Workaround

Browser console (replace `{id:1}` with the actual user ID):
```js
fetch("/api/auth/users",{method:"DELETE",credentials:"include",headers:{"Content-Type":"application/json"},body:JSON.stringify({id:1})}).then(function(r){return r.json()}).then(console.log)
```

## Environment

- Mission Control v2.0.1 (standalone)
- Chromium 135, Edge, Firefox

---
*This issue was filed by an AI assistant (OpenClaw agent) on behalf of a user during deployment troubleshooting. The technical details were verified in a live environment.*

---
*This issue was filed by an AI assistant (OpenClaw agent) on behalf of a user during deployment troubleshooting. The technical details were verified in a live environment.*

nmsn:
## Fixes #614

Delete user from UI fails with 'Request body required'

**Root cause**: `handleDelete` sent user ID as query param (`?id=xxx`) but the backend API route requires it in the request body.

**Change**: Move `id` from query parameter to JSON body.

### DIFF
--- a/src/components/panels/user-management-panel.tsx
+++ b/src/components/panels/user-management-panel.tsx
@@ -164,7 +164,11 @@ export function UserManagementPanel() {
   const handleDelete = async (u: UserRecord) => {
     if (u.id === currentUser?.id) return
     try {
-      const res = await fetch(`/api/auth/users?id=${u.id}`, { method: 'DELETE' })
+      const res = await fetch('/api/auth/users', {
+        method: 'DELETE',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify({ id: u.id }),
+      })
       const data = await res.json().catch(() => ({}))
       if (res.ok) {
         showFeedback(true, t('deletedUser', { username: u.username }))


0xbrainkid (approved):
Approved: small client fix matches the DELETE route contract (JSON body with id) and preserves self-delete guard handling.

0xNyk (approved):
Approving — single-component fix. `handleDelete` now passes the user id in a JSON body with `Content-Type: application/json` instead of as a query param, matching what the API route expects. Closes #614.

PR #633, opened 2026-04-29
Self-contained Docker stack + /chat fixes for shared host claude sessionsFeat/docker stack

nnnet:
## Summary

Makes Mission Control runnable as a single self-contained Docker container that drives the operator's host-installed Claude Code / Codex / OpenCode CLIs without requiring re-authentication inside the container, and without an OpenClaw gateway (which is macOS-only and not available on Linux hosts). Along the way, fixes a CSP/hydration regression that broke `/chat`, and several issues in the local-Claude-sessions pipeline.

This is a focused contribution on the **operator path** (single user running MC locally on a Linux/Docker host alongside their own claude CLI) — not on the multi-tenant or production deployment story.

## Why

I wanted to use the existing `/chat` page to drive the same Claude Code session I have open in a host terminal, while running MC itself in Docker. That setup hits a few rough edges:

- The CLI binaries need to come from the host (the operator is already authenticated), but the container needs to be able to spawn them.
- `/chat` was not interactive in my browser — strict CSP blocked hydration entirely.
- `claude --resume <id>` from inside the container couldn't find any session, even though it could see the jsonl files from the bind-mounted host `~/.claude`.
- The "Active" session list kept growing forever and surfacing dead jsonls as live.

This PR addresses each of those.

### 1. `feat(docker): self-contained docker stack with host config projection`

- `Dockerfile`: bake `claude` and `codex` into the image as a fallback (host's `~/.local/bin` still wins via `PATH` order); add `tmux` and `jq` (required by `/chat` PTY view and runtime probes); reuse the slim image's existing uid 1000 user as `nextjs` so bind-mounted host files (typical Linux uid 1000) are read/written without chown.
- `docker-compose.yml`: run as `1000:1000`; project the host user's `\$HOME` (`.local/bin`, `.bun`, `.claude`, `.claude.json`, `.local/share/claude`) into the container; bind-mount `/mnt` and `\${HOME}` at the same absolute path so paths are identical inside and outside; bump memory limit to 2G (Next.js 16 + node-pty + the task-dispatch loop OOM at the upstream 512M when `/chat` opens a terminal); pass `ANTHROPIC_API_KEY` and `OPENAI_API_KEY` from `.env` so direct-API dispatch works without a gateway; add `MC_HOST_SESSION_MODE` env (see commit 3).
- `Makefile` (new): single entrypoint for the operator — `up`, `down`, `restart`, `recreate`, `build`, `rebuild` (no-cache), `ps`, `logs`, `shell`, `status`, `wait-ready`, `reset-db`, `nuke`. URL fixed to `http://127.0.0.1:7012`.
- `.gitignore`: ignore `WALKTHROUGH.md` (operator-local notes file kept on disk, never committed).

### 2. `fix(csp): pass nonce to next-themes ThemeProvider so hydration works`

The strict CSP set in `src/proxy.ts` uses `script-src 'nonce-X' 'strict-dynamic'` which blocks every inline `<script>` that does not carry the matching nonce. `next-themes` injects an inline anti-FOUC script at render time; without an explicit `nonce` prop it ships that script with no nonce attribute, so the browser blocks it. With `strict-dynamic`, that boot script is the trust root for everything else: when the browser blocks it, **no further scripts can load**, so React never hydrates. Symptoms: chat input has no working Send button, fetch from client code throws `TypeError: Failed to fetch`, etc.

Fix: pass the per-request nonce (already available in layout via the `x-nonce` request header that `proxy.ts` sets) into `<ThemeProvider nonce={nonce} ...>`. `next-themes >= 0.4` propagates this to its inline tag.

I also left two `console.log('[DEBUG csp] ...')` lines commented out in `proxy.ts` and `layout.tsx` — useful for the next person who has to debug nonce flow end-to-end. Happy to remove them entirely if reviewers prefer.

### 3. `fix(chat): /api/sessions/continue works on shared host claude sessions`

Four sub-issues collapsed into this commit:

- **Wrong cwd**: `claude --resume <id>` only finds the transcript if process `cwd` matches the project path encoded in `~/.claude/projects/<encoded>/`. MC's `runCommand` defaulted to `/app`, so resume silently picked up no conversation.
- **Lossy decoding**: claude collapses both `/` and `_` to `-` in the encoded directory name, so `/foo/bar_baz` and `/foo/bar/baz` both round-trip to `-foo-bar-baz`. Naive decoding produced wrong cwds → `cd` failed → claude never ran. The fix reads the authoritative `cwd` field directly from the first JSON line of the session jsonl.
- **`spawn claude ENOENT` under Next 16 standalone**: a direct `child_process.spawn('claude', ...)` from the Next.js standalone server consistently produced `ENOENT` even though the binary was reachable via `which claude` and via `node -e 'spawn(...)'` in the same container with the same env. The exact root cause was not pinned down; I worked around it by resolving the absolute path to `claude` via `fs.access(X_OK)` over `PATH`, and then spawning through `sh -c \"cd <cwd> && exec <bin> --print --resume <id>\"` instead of direct execvp. `shQuote()` keeps both paths safe.
- **Race policy**: when MC `--resume`s a session that already has a live host CLI, both processes append to the same jsonl. Added env `MC_HOST_SESSION_MODE`:
  - `coexist` (default) — both write, each picks up the other's writes on its next prompt.
  - `block-active` — return 409 if jsonl mtime < 60s ago (the host CLI looks live).
  - `nudge` — coexist + best-effort \`utimes()\` after the reply.

A `TODO` marker is added for the proper UX-level fix to the long Send→reply latency: replace the blocking `claude --print` with an SSE endpoint backed by `--output-format stream-json` so the chat UI can render tokens incrementally. Out of scope for this PR.

### 4. `fix(chat): tighter active-session window + remove orphan rows`

Two related issues caused the `/chat` session list to drift from reality:

- The 90-minute "active" window flagged any jsonl touched in the last hour and a half as live. For an operator with several claude sessions across projects this surfaced almost everything as active, including ones they had finished with hours of think-time ago. Tightened to 15 minutes in both the scanner-side (`ACTIVE_THRESHOLD_MS`) and the API-side derived recovery (`LOCAL_SESSION_ACTIVE_WINDOW_MS`) so they stay coherent.
- Rows in `claude_sessions` lived forever even after the underlying jsonl was deleted (manual cleanup, project rename, or `claude --resume` that creates a new id). The API still surfaced those orphans as "active" off the stale `last_message_at`. Now, after a scan cycle, the scanner also `DELETE`s rows whose `session_id` is not in the freshly-scanned set, and reports `removed N orphan(s)` when this happens.

### 5. `fix(chat): keep input anchored, reply renders into transcript`

- Stop excluding `session:*` conversations from the polling fallback — without this, `/chat` would freeze until the operator hit F5 every time SSE dropped. Cadence is now parameterized via `NEXT_PUBLIC_CHAT_POLL_INTERVAL_MS` (1500ms component default; 1000ms in docker-compose) so it can be tuned without code changes.
- Drop the standalone `lastReply` panel that used to show the most recent assistant message above the input. It duplicated the transcript and, on long replies, pushed the input field below the viewport. The reply now lands in the transcript via `onRefreshTranscript()` — same `SessionMessage` rendering, same formatting, same scroll behaviour as the rest of the conversation. The prompt input row stays anchored to the bottom regardless of reply size, matching how the host claude CLI itself displays things.

## What I deliberately did **not** change

- The OpenClaw gateway path. macOS-only OpenClaw is untouched; this PR only adds a parallel direct-API dispatch path that already existed in the codebase (`callClaudeDirectly`) — it doesn't modify it.
- Multi-provider direct dispatch (OpenAI, local LMStudio). I prototyped this on a separate branch but pulled it back out of this PR — it's a separate feature and should be its own review.
- Streaming reply (SSE / `--output-format stream-json`). Marked as `TODO` in `route.ts`. Worth a follow-up; the current 1000ms polling is a workable but not great UX for long replies.

## Test plan

- [ ] `make up` brings the container up and reaches HTTP 200 at `http://127.0.0.1:7012/login` within 30s.
- [ ] `/setup` creates an admin and `/login` works.
- [ ] `/chat` loads without CSP errors in DevTools Console; the Send button is interactive.
- [ ] In `/chat`, sending a prompt to a host-created Claude Code session:
  - returns 200 from `POST /api/sessions/continue`
  - the user prompt and the assistant reply both appear in the host session jsonl
  - both also appear in the MC transcript view (via the polling refresh)
  - the input row stays anchored to the bottom, regardless of reply size
- [ ] Setting `MC_HOST_SESSION_MODE=block-active` in `.env` and pinging an actively-written jsonl returns HTTP 409 with the "Wait for it to go idle" message.
- [ ] After deleting some jsonls under `~/.claude/projects/<some-dir>/`, the next sync removes the corresponding rows from `claude_sessions` and the `/chat` Active list shrinks accordingly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,6 @@ playwright-report/
 # Claude Code context files (root CLAUDE.md is committed for AI agent discovery)
 **/CLAUDE.md
 !/CLAUDE.md
+
+# Local agent walkthrough (kept on disk, not committed)
+WALKTHROUGH.md
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,8 +32,20 @@ WORKDIR /app
 ENV NODE_ENV=production
 # curl, CA certs, python3, git needed for agent runtime installers (OpenClaw, Hermes)
 # procps provides `ps` and `uptime` used by system-monitor APIs
-RUN apt-get update && apt-get install -y curl ca-certificates python3 git make g++ procps --no-install-recommends && rm -rf /var/lib/apt/lists/*
-RUN addgroup --system --gid 1001 nodejs && adduser --system --uid 1001 nextjs
+RUN apt-get update && apt-get install -y curl ca-certificates python3 git make g++ procps tmux jq --no-install-recommends && rm -rf /var/lib/apt/lists/*
+
+# Bake Claude Code + Codex CLIs into the image as a fallback so the
+# Settings → Agent Runtimes panel reports "Installed" even before the host
+# bind-mounts (compose adds ${HOME}/.local/bin at runtime, which takes
+# precedence in PATH and provides authenticated host binaries).
+RUN npm install -g @anthropic-ai/claude-code @openai/codex 2>&1 | tail -5
+
+# node:22-slim already ships a `node` user at uid 1000; reuse it as our
+# `nextjs` alias so bind-mounted host files (typical Linux uid 1000) read directly.
+RUN if ! id -u nextjs >/dev/null 2>&1; then \
+      usermod --login nextjs --move-home --home /home/nextjs node && \
+      groupmod --new-name nodejs node ; \
+    fi
 COPY --from=build /app/.next/standalone ./
 COPY --from=build /app/.next/static ./.next/static
 COPY --from=build /app/public ./public
@@ -46,7 +58,8 @@ RUN mkdir -p .data && chown nextjs:nodejs .data
 RUN echo 'const http=require("http");const r=http.get("http://localhost:"+(process.env.PORT||3000)+"/api/status?action=health",s=>{process.exit(s.statusCode===200?0:1)});r.on("error",()=>process.exit(1));r.setTimeout(4000,()=>{r.destroy();process.exit(1)})' > /app/healthcheck.js
 COPY docker-entrypoint.sh /app/docker-entrypoint.sh
 RUN chmod 755 /app/docker-entrypoint.sh && \
-    chmod -R a+rX /app/public/ /app/src/
+    chmod -R a+rX /app/public/ /app/src/ && \
+    chown -R nextjs:nodejs /app /home/nextjs
 USER nextjs
 ENV PORT=3000
 EXPOSE 3000
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,118 @@
+# Mission Control — local stack control plane.
+# Use `make help` for the full target list.
+
+SHELL := /bin/bash
+.ONESHELL:
+.SHELLFLAGS := -eu -o pipefail -c
+
+PROJECT_DIR := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
+COMPOSE     := docker compose
+CONTAINER   := mission-control
+URL         := http://127.0.0.1:7012
+
+.DEFAULT_GOAL := help
+
+# ── Help ───────────────────────────────────────────────────────────────────
+.PHONY: help
+help:  ## List targets
+	@awk 'BEGIN{FS=":.*##"; printf "\nUsage: make \033[36m<target>\033[0m\n\nTargets:\n"} \
+	      /^[a-zA-Z0-9_-]+:.*##/ { printf "  \033[36m%-22s\033[0m %s\n", $$1, $$2 }' \
+	      $(MAKEFILE_LIST)
+
+# ── Compose lifecycle ──────────────────────────────────────────────────────
+.PHONY: up
+up:  ## Bring stack up and wait for /login to respond
+	@cd $(PROJECT_DIR)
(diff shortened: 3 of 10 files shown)


0xbrainkid:
REWORK: this is not safe to merge as-is.

Reason: `docker-compose.yml` bind-mounts the host home directory and credential/config directories into the app container with read-write access (`${HOME}:${HOME}:rw`, `${HOME}/.claude`, `${HOME}/.claude.json`, `${HOME}/.local/bin`, etc.). That gives the web app/agent runtime broad write access to host files and secrets, which is too large a security boundary change for a default Docker stack.

Next step: make the host mounts opt-in behind an explicit compose override/profile (or mount the minimum required paths read-only by default), document the risk, and keep the default stack least-privilege. Then rerun validation.

PR #635, opened 2026-05-01
Add Mission Control architecture, Passive Income Bot, memory UI, and routing validation

niko4244:
## Summary

This PR packages the current Mission Control checkpoint stack from `niko/main`.

### Included

- Added Mission Control architecture documentation:
  - agent registry
  - skill registry
  - schedules
  - evidence log schema
  - risk and approval policy
  - bot output templates
  - overall architecture spec

- Added Passive Income Bot v1/v1.1 foundation:
  - bot script
  - API route
  - server wrapper
  - test coverage
  - draft-only / approval-gated behavior

- Added Mission Control Memory UI and review-route foundation:
  - memory API routes
  - memory page/UI
  - supporting server services
  - seed and validation scripts

- Added routing validation foundation:
  - MCP-style command classification
  - dangerous-operation validation
  - stricter handling for unknown commands
  - shell allowlist checks

### Checks Passed

- `pnpm typecheck`
- `pnpm lint`
- `pnpm test`
- `pnpm build`

### Push Note

Direct push to `origin/main` failed due to GitHub permissions on `builderz-labs/mission-control` for account `niko4244`, so the validated branch was pushed to the fork remote `niko` instead.

PR #637, opened 2026-05-01
Add Systems Curator v1

niko4244:
# Summary
Describe what changed and why.

# Risk Level
Low / Medium / High (pick one)

# Tests
List commands run and results.

# Contribution Checklist
- [ ] Lint/typecheck/build passing
- [ ] Security review done if auth/data/crypto touched

# Notes
Anything reviewers should know.

### DIFF
--- a/scripts/systems-curator.cjs
+++ b/scripts/systems-curator.cjs
@@ -1,0 +1,273 @@
+#!/usr/bin/env node
+
+const fs = require('node:fs');
+const path = require('node:path');
+const { execFileSync } = require('node:child_process');
+
+const REPO_ROOT = path.resolve(__dirname, '..');
+
+const EXPECTED_DOCS = [
+  'docs/mission-control/MISSION_CONTROL_ARCHITECTURE.md',
+  'docs/mission-control/AGENT_REGISTRY.md',
+  'docs/mission-control/SKILL_REGISTRY.md',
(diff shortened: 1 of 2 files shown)


0xbrainkid:
REWORK: this clean branch is small, but the included test is not mergeable against current `main`.

I checked the diff (2 files) and ran the new CLI on the PR head. The script reports `status: "FAIL"` on a clean checkout because all of the Mission Control docs/bot/memory files it expects are absent from current `main`. More importantly, `src/lib/__tests__/systems-curator.test.ts` asserts `docs_missing` is `[]`, so this PR depends on files from the larger Mission Control stack and will fail once dependencies are installed and the test runs.

Next step: either (a) base this PR on the branch that adds those expected Mission Control files, or (b) change Systems Curator v1/tests so they treat those files as optional/missing on current `main` and assert the current expected FAIL/WARN output instead of `docs_missing: []`.

0xNyk:
Thanks — the systems-curator script alone is the most reviewable piece in your stack of #635-#641, and the 2-file scope is the right shape.

Closing this one because the PR description is the unfilled template, but **please reopen** it as a fresh PR with:

1. A written summary explaining what `systems-curator.cjs` does
2. Why MC needs it (use case, problem it solves)
3. Risk level + what tests cover the new behavior
4. Confirmation it's branched off the latest `main`

That's the version I'd merge.

PR #642, opened 2026-05-02
fix: preserve gateway auth for saved urls

tl2811:
## Summary\n- keep saved browser gateway URLs on the server-token connect path instead of connecting without auth\n- set standalone runtime data dir to the repo .data directory so production uses the real gateway DB\n\n## Verification\n- pnpm build\n- systemctl restart mission-control.service\n- /api/gateways/connect returns wss://gw.doremon.app with token_set=true\n- https://mc.doremon.app/login returns 200\n

### DIFF
--- a/scripts/start-standalone.sh
+++ b/scripts/start-standalone.sh
@@ -40,6 +40,8 @@ if [[ -f "$PROJECT_ROOT/.env" ]]; then
   set +a
 fi
 
+export MISSION_CONTROL_DATA_DIR="${MISSION_CONTROL_DATA_DIR:-$PROJECT_ROOT/.data}"
+
 # Next.js standalone server reads HOSTNAME to decide bind address.
 # Default to 0.0.0.0 so the server is accessible from outside the host.
 export HOSTNAME="${HOSTNAME:-0.0.0.0}"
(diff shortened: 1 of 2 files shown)


0xbrainkid (approved):
Reviewed the saved gateway auth path and standalone data-dir default. The startup flow now obtains the server-side gateway token before connecting to a saved browser URL, and the standalone launcher defaults data storage back to the repo .data directory while preserving any explicit MISSION_CONTROL_DATA_DIR override. No blocking issues found. Note: GitHub reports no checks configured for this branch.

0xNyk (approved):
Approving — small but auth-relevant. Two fixes:

1. **Auth preservation on saved URLs.** Previously `connect(localGatewayUrl)` was called directly when a user had a saved gateway URL, bypassing the gateway's `/connect` token-issuing endpoint and leaving the WebSocket without a token. The new path tries `connectWithPrimaryGateway(localGatewayUrl)` first — which fetches a token via `/api/gateways/.../connect`, then connects to the user's saved URL with that token. Falls back to the original direct connect if the gateway didn't issue a token. This preserves auth where it was being silently dropped.

2. **`MISSION_CONTROL_DATA_DIR` in `start-standalone.sh`.** Defaults the env to `PROJECT_ROOT/.data` so the standalone server points at the real gateway DB instead of whatever the runtime CWD happens to be.

Walked the auth flow end-to-end. Merging.

PR #643, opened 2026-05-03
feat(docker): expose NEXT_PUBLIC_* as build args for browser-side gateway config

YoungSella:
## Problem

When deploying Mission Control via Docker behind a reverse proxy (e.g. nginx-proxy-manager / Caddy / Traefik) where the OpenClaw gateway is exposed on a **different subdomain or hostname** than the dashboard itself, there is currently no way to configure the browser-side WebSocket URL without forking and rebuilding the image manually.

## Root Cause

`src/app/[[...panel]]/page.tsx:182-190` reads:

```ts
const explicitWsUrl = localGatewayUrl || process.env.NEXT_PUBLIC_GATEWAY_URL || ''
const gatewayPort   = process.env.NEXT_PUBLIC_GATEWAY_PORT || '18789'
const gatewayHost   = process.env.NEXT_PUBLIC_GATEWAY_HOST || window.location.hostname
const protocol      = process.env.NEXT_PUBLIC_GATEWAY_PROTOCOL || ...
```

These `NEXT_PUBLIC_*` variables are baked into the client bundle at **build time** (`pnpm build`). However the published `Dockerfile` does not declare them as `ARG`s in the `build` stage, so passing `--build-arg NEXT_PUBLIC_GATEWAY_URL=...` to `docker build` (or via `args:` in `compose.yml`) has no effect — the bundle is always built with defaults, and the browser falls back to `wss://${window.location.hostname}:18789/gateway-ws`.

For deployments where the dashboard sits at e.g. `https://mc.example.com` and the gateway sits at `https://openclaw.example.com` (proxied by nginx to `127.0.0.1:18789` with WS upgrade), this default is unreachable from the browser.

The `.env.example` already documents these variables clearly, but operators have no clean path to inject them into the build.

## Fix

Add `ARG` and `ENV` declarations for all browser-facing `NEXT_PUBLIC_*` variables in the `build` stage of the `Dockerfile`, immediately before `RUN pnpm build`. This is a backward-compatible change — the defaults remain empty so behaviour is identical when no build args are passed, but operators can now configure the build via:

```yaml
# compose.yml
services:
  mission-control:
    build:
      context: .
      args:
        NEXT_PUBLIC_GATEWAY_URL: wss://openclaw.example.com
        NEXT_PUBLIC_GATEWAY_CLIENT_ID: openclaw-control-ui
```

Variables added (matching `.env.example`):

- `NEXT_PUBLIC_GATEWAY_URL`
- `NEXT_PUBLIC_GATEWAY_HOST`
- `NEXT_PUBLIC_GATEWAY_PORT`
- `NEXT_PUBLIC_GATEWAY_PROTOCOL`
- `NEXT_PUBLIC_GATEWAY_REVERSE_PROXY`
- `NEXT_PUBLIC_GATEWAY_CLIENT_ID`
- `NEXT_PUBLIC_GATEWAY_OPTIONAL`
- `NEXT_PUBLIC_COORDINATOR_AGENT`
- `NEXT_PUBLIC_GOOGLE_CLIENT_ID`

## Test Plan

1. Build with: `docker build --build-arg NEXT_PUBLIC_GATEWAY_URL=wss://example.com .`
2. Run image, open dashboard in browser
3. DevTools → Network → WS → confirm WS connects to `wss://example.com/gateway-ws` (instead of `wss://${page-host}:18789/gateway-ws`)
4. Build without build args → unchanged behaviour (defaults to current logic)

## Discovered via

Project EIGHTBALL — multi-agent self-hosted deployment on Mac mini M4 with separate subdomains for gateway (`openclaw.example.io`) and dashboard (`mc.example.io`), both behind nginx-proxy-manager with Let's Encrypt SSL.

## Backwards Compatibility

100% backward compatible. No defaults change. Operators who don't pass build args get exactly the current behaviour.

## Optional Follow-up

Could add a `Deployment Modes / Reverse Proxy` section to README pointing to this approach, since `.env.example` already covers the variable semantics but doesn't mention the build-arg path.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,33 @@ RUN if [ -f pnpm-lock.yaml ]; then \
 FROM base AS build
 COPY --from=deps /app/node_modules ./node_modules
 COPY . .
+
+# ─── PR-CANDIDATE: NEXT_PUBLIC_* baked into client bundle ──────────────────
+# Next.js inlines NEXT_PUBLIC_* into the client JS at build time. Without
+# these ARG/ENV pairs, downstream operators (Docker / CI / k8s) cannot
+# configure the gateway URL for the browser without a custom image build.
+# Discovered via Project EIGHTBALL deployment with separate subdomain for
+# the gateway (openclaw-mac.example.io) vs dashboard (mc-mac.example.io).
+ARG NEXT_PUBLIC_GATEWAY_URL=
+ARG NEXT_PUBLIC_GATEWAY_HOST=
+ARG NEXT_PUBLIC_GATEWAY_PORT=
+ARG NEXT_PUBLIC_GATEWAY_PROTOCOL=
+ARG NEXT_PUBLIC_GATEWAY_REVERSE_PROXY=
+ARG NEXT_PUBLIC_GATEWAY_CLIENT_ID=
+ARG NEXT_PUBLIC_GATEWAY_OPTIONAL=
+ARG NEXT_PUBLIC_COORDINATOR_AGENT=
+ARG NEXT_PUBLIC_GOOGLE_CLIENT_ID=
+ENV NEXT_PUBLIC_GATEWAY_URL=${NEXT_PUBLIC_GATEWAY_URL}
+ENV NEXT_PUBLIC_GATEWAY_HOST=${NEXT_PUBLIC_GATEWAY_HOST}
+ENV NEXT_PUBLIC_GATEWAY_PORT=${NEXT_PUBLIC_GATEWAY_PORT}
+ENV NEXT_PUBLIC_GATEWAY_PROTOCOL=${NEXT_PUBLIC_GATEWAY_PROTOCOL}
+ENV NEXT_PUBLIC_GATEWAY_REVERSE_PROXY=${NEXT_PUBLIC_GATEWAY_REVERSE_PROXY}
+ENV NEXT_PUBLIC_GATEWAY_CLIENT_ID=${NEXT_PUBLIC_GATEWAY_CLIENT_ID}
+ENV NEXT_PUBLIC_GATEWAY_OPTIONAL=${NEXT_PUBLIC_GATEWAY_OPTIONAL}
+ENV NEXT_PUBLIC_COORDINATOR_AGENT=${NEXT_PUBLIC_COORDINATOR_AGENT}
+ENV NEXT_PUBLIC_GOOGLE_CLIENT_ID=${NEXT_PUBLIC_GOOGLE_CLIENT_ID}
+# ────────────────────────────────────────────────────────────────────────────
+
 RUN pnpm build
 
 FROM node:22.22.0-slim AS runtime


0xbrainkid (approved):
Reviewed the Dockerfile change. This safely exposes the documented browser-facing NEXT_PUBLIC_* values as build args before pnpm build, preserving existing defaults when args are omitted. No blocking issues found. Note: GitHub reports no checks configured for this branch.

YoungSella:
Quick follow-up — wanted to confirm this patch has been running in production on Project EIGHTBALL since the day the PR was opened (2 May 2026). The Mission Control image built with these build args (mission-control:eightball tag, NPM-proxied subdomain mission-control-mac.sella-entertainment.io with separate gateway subdomain openclaw-mac.sella-entertainment.io) is connecting to the gateway via wss://openclaw-mac.../gateway-ws exactly as expected. WebSocket handshake succeeds, browser sees the right host, no :18789 port leakage in the URL.
@0xNyk — friendly ping for codeowner review when you get a chance. Happy to add tests, documentation, or split the change differently if useful. Also happy to write a follow-up README section under "Deployment Modes / Reverse Proxy" if that would help downstream operators.
Two small observations from running 2026.5.3 with this build that might be unrelated but worth flagging — happy to open separate issues if you'd like:

agents.defaults.models is a dict and the first key gets picked as the chat-default if agents.defaults.model is unset. In a default install where nomic-embed-text:latest happens to be first, this surfaces as 400 {"error":"\"nomic-embed-text:latest\" does not support chat"}. Could either be solved by filtering for chat-capable models or by emitting a warning when an embedding model is selected as the chat default.
Web chat UI shows internal [Core execution] and [OpenClaw heartbeat poll] messages as if they were user/assistant turns. Cosmetic but noisy — operators new to the project will think they're seeing chat history.

Thanks for building this — the multi-agent architecture is exactly what I needed for my self-hosted setup.

0xbrainkid (approved):
$PR queue gate re-check: approved. Re-reviewed the current Dockerfile-only diff after latest update: it only adds documented NEXT_PUBLIC_* build args/env values before pnpm build, which matches Next.js client-side env behavior and preserves default empty behavior when args are omitted. No checks are configured on the branch. No merge performed.

0xNyk (approved):
Approving — exposes `NEXT_PUBLIC_*` gateway-config keys as Docker build ARGs/ENVs so operators can configure the **browser-side** WebSocket URL (subdomain split between dashboard and gateway) without forking the image.

**Build-time env audit:**
- All exposed keys are `NEXT_PUBLIC_*` and named for client-side gateway config (`GATEWAY_URL`, `GATEWAY_HOST`, `GATEWAY_PORT`, `GATEWAY_PROTOCOL`, `GATEWAY_REVERSE_PROXY`, `GATEWAY_CLIENT_ID`, `GATEWAY_OPTIONAL`, `COORDINATOR_AGENT`, `GOOGLE_CLIENT_ID`).
- None are server secrets — `GOOGLE_CLIENT_ID` is a public OAuth client identifier, the gateway URL/host/port/protocol are intended-public connection coordinates, the rest are UI configuration.
- The `NEXT_PUBLIC_` prefix means Next.js will inline these into the client bundle anyway when set — this PR just exposes them at the right point in the build pipeline.

**No** `NEXT_PUBLIC_*_SECRET` / `NEXT_PUBLIC_*_KEY` style names slipped in. No `ARG` for `ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, `OPENCLAW_GATEWAY_TOKEN`, etc. Clean.

YoungSella:
Thanks @0xNyk for the merge and @0xbrainkid for the thorough review — really appreciate the quick turnaround on both passes. 🙏

PR #647, opened 2026-05-05
fix(chat,csp): nonce hydration + chat session continuity with host Claude CLI

nnnet:
## Summary

Four small, independent bug fixes:

1. **CSP nonce → next-themes hydration.** The CSP nonce was not forwarded to `next-themes`' `ThemeProvider`, so theme hydration was blocked under our default CSP and the dashboard rendered with the wrong theme on the first paint. `src/proxy.ts` now propagates the nonce; `src/app/layout.tsx` consumes it.

2. **`/api/sessions/continue` works on shared host Claude sessions.** When operators bind-mount their host `~/.claude/projects` into the container (a common pattern with the host-config projection in PR #646), the resolver couldn't open sessions whose `.jsonl` lived under that mount. Path resolver now accepts both container-local and host-projected paths.

3. **Tighter active-session window + drop orphan rows.** MC was treating any session whose `.jsonl` was touched recently as "alive" — even after the underlying Claude process had long stopped. The window is tighter now and orphan rows are filtered.

4. **Chat input stays anchored.** The chat input lost focus on every message because the whole transcript re-rendered. We anchor the input and patch the transcript instead.

These are user-visible defects that anyone running MC behind a CSP or sharing host Claude state hits within minutes of using the chat panel. None depend on each other or on any other in-flight PR.

## Test plan

- Visit the dashboard with a strict CSP — theme matches the user's preference on first paint, no hydration errors in console.
- Bind-mount host `~/.claude/projects` into the dev container, open a session that exists only on the host — `/api/sessions/continue` resolves it.
- Watch `/api/sessions` after killing a Claude process — the orphan row disappears within the new window.
- Send several chat messages — input keeps focus across each reply.

## Provenance

Squashes our fork's commits:
- `9ac39a6 fix(csp): pass nonce to next-themes ThemeProvider so hydration works`
- `d7c10b2 fix(chat): /api/sessions/continue works on shared host claude sessions`
- `8463a70 fix(chat): tighter active-session window + remove orphan rows`
- `2fce07a fix(chat): keep input anchored, reply renders into transcript`

### DIFF
--- a/src/app/api/sessions/continue/route.ts
+++ b/src/app/api/sessions/continue/route.ts
@@ -1,20 +1,172 @@
-import { promises as fs } from 'node:fs'
+import { promises as fs, constants as fsConstants } from 'node:fs'
 import path from 'node:path'
+import os from 'node:os'
 import { NextRequest, NextResponse } from 'next/server'
 import { requireRole } from '@/lib/auth'
 import { logger } from '@/lib/logger'
 import { runCommand } from '@/lib/command'
 import { getOpenCodeExecutable } from '@/lib/opencode-sessions'
 
+/**
+ * Resolve a CLI binary to an absolute path by scanning PATH directories.
+ * Next.js standalone server's process.env may not always allow Node's
+ * default execvp lookup to find tools installed in user-local bins
+ * (`~/.local/bin`) — observed empirically as `spawn claude ENOENT` even
+ * though `which claude` succeeds in the same container. Resolving the
+ * absolute path eliminates the ambiguity. Falls back to bare name.
+ */
+async function resolveExecutable(name: string): Promise<string> {
+  if (name.includes('/')) return name
+  const candidates = [
+    process.env.CLAUDE_BIN,
+    `/home/nextjs/.local/bin/${name}`,
+    `/usr/local/bin/${name}`,
+    `/usr/bin/${name}`,
+  ].filter((p): p is string => !!p && p.endsWith(`/${name}`))
+  const pathDirs = (process.env.PATH || '').split(':').filter(Boolean)
+  for (const dir of pathDirs) candidates.push(path.join(dir, name))
+  for (const candidate of candidates) {
+    try {
+      await fs.access(candidate, fsConstants.X_OK)
+      return candidate
+    } catch {
+      continue
+    }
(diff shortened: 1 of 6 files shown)


0xbrainkid (approved):
Reviewed the current diff. The nonce propagation, session transcript refresh behavior, and Claude resume cwd/bin resolution all look safe to land. I also checked the session sync cleanup path; no blocking issue found.\n\nGates run:\n- `pnpm exec vitest run src/app/api/sessions/__tests__/continue-route-opencode.test.ts src/app/api/sessions/__tests__/transcript-opencode.test.ts` ✅\n- `pnpm exec tsc --noEmit` was attempted; it still fails on pre-existing unrelated project-wide errors (nav-rail Button props / missing next-intl types), with no new errors identified in the touched files.

0xNyk (approved):
Approving — four small fixes, each scoped to a clear correctness problem. Walked the security-relevant paths:

1. **CSP nonce → `next-themes` hydration.** `src/proxy.ts` propagates the existing nonce; `src/app/layout.tsx` consumes it and passes `nonce={nonce}` to `ThemeProvider`. No new `unsafe-inline` or `script-src` permissions added — this is a *correctness* fix that wires the existing nonce through to a component that needed it. The commented-out `[DEBUG csp]` console logs are fine to retain as is.

2. **`/api/sessions/continue` cwd resolution.** Reads the jsonl transcript header, parses lines until it finds a `cwd` that starts with `/`. The authoritative `cwd` field comes from the operator-controlled `~/.claude/projects/<id>.jsonl` — same trust boundary as the rest of MC's claude-session integration. Not a new attack surface.

3. **`resolveExecutable()` PATH walking.** Hardcoded fallback list (`CLAUDE_BIN` env, `/home/nextjs/.local/bin`, `/usr/local/bin`, `/usr/bin`, then `PATH` dirs). No user input touches the resolution. Fine.

4. **Orphan `claude_sessions` row cleanup.** Pure DB hygiene — deletes rows whose jsonl no longer exists. The `derivedActive` mtime fallback was previously reporting deleted sessions as 'Active'.

Plan agent flagged this as Tier B for CSP-scope concern; verified that's not the case here. Merging.

Issue #608, opened 2026-04-22
[Bug] Gateway integration broken in distributed/containerized deployments — CLI subprocess unavailable in MC container


jmmc-tools:
## Describe the Bug

When Mission Control is deployed in a **containerized or distributed environment** (MC and openclaw gateway running in separate pods/containers), all gateway-based task dispatch and aegis review calls silently fail. Tasks accumulate `dispatch_attempts` indefinitely but never leave `assigned` status — no LLM response is ever received or stored.

Two independent bugs combine to cause this:

**Bug 1 — CLI binary not available in MC container (`ENOENT`)**
`callOpenClawGateway` and both dispatch paths in `task-dispatch.ts` shell out to the `openclaw` CLI binary via `runOpenClaw(['gateway', 'call', ...])`. In a containerized MC deployment, this binary does not exist in the MC image. Every gateway call throws `spawn openclaw ENOENT`.

**Bug 2 — `agent.wait` RPC never returns LLM text (`deliver: false`)**
Even when the CLI is available, both task dispatch and aegis review set `deliver: false` in the agent invocation params. The `agent.wait` WS response contains only lifecycle metadata (`{runId, status, startedAt, endedAt}`) — the actual LLM response text is never included. Every dispatch ends with `"Agent returned empty response"` and the task never progresses.

## Steps to Reproduce

1. Deploy Mission Control as a standalone Docker container or Kubernetes pod
2. Deploy openclaw gateway as a **separate** container/service (different host or pod)
3. Configure `OPENCLAW_GATEWAY_HOST` and `OPENCLAW_GATEWAY_PORT` environment variables in MC to point to the gateway over the network
4. Register an agent in MC with `openclawId` matching a gateway-registered agent
5. Create a task and assign it to that agent
6. Observe: task never leaves `assigned` status, `dispatch_attempts` increments each cycle

**MC logs show either:**
```
Error: spawn openclaw ENOENT
```
or (if CLI is somehow present):
```
Task dispatch failed  ERR: Agent returned empty response
PAYLOAD: {"runId":"task-dispatch-1-...","status":"ok","startedAt":...,"endedAt":...}
```

## Expected Behavior

MC dispatches the task to the openclaw gateway over the network, receives the full LLM response text, stores it as a task comment, and transitions the task to `in_progress` → `review` → `done`.

## Screenshots

N/A — bug is observable via MC logs and DB task status.

## Environment

- OS: Linux (Kubernetes pod, `linux/amd64`)
[...]

Issue #609, opened 2026-04-22
[Feature] Add OpenAI-compatible API endpoint as an LLM provider


bsarkisov:
## Problem Statement
Current setup allows either use cloud-based models or ollama provider located on the same server as mission control. 
This doesn't allow user use any locally-deployed LLMs (QWEN, Gemma, Mistral etc.) to power Mission Control if they're deployed on a separate server (as it usually is).
So it forces user to use ollama instead of llama/vllm and other available engines.

## Proposed Solution
Add OpenAI-compatible API endpoint to support onther LLM serving engines.

0xNyk:
Resolved by **#648** which just merged. Direct multi-provider dispatch routes `local/*` / `ollama/*` / `lmstudio/*` / `litellm/*` model prefixes to a configurable OpenAI-compatible endpoint via `LOCAL_LLM_ENDPOINT` (with optional `LOCAL_LLM_API_KEY`).

Works with any OpenAI `/v1/chat/completions` server: vLLM, llama.cpp server, LMStudio (default), Ollama, or a liteLLM proxy fronting multiple backends. From the deployment doc:

```env
LOCAL_LLM_ENDPOINT=http://your-vllm-host:8000/v1
LOCAL_LLM_API_KEY=optional-if-your-proxy-needs-it
```

Closing.

Issue #588, opened 2026-04-13
[Bug] error  Spawn openclaw Enoent


KianaKaslanaWaifu:
## Describe the Bug
i am using windows  i tried to create an agent and got this 

## Steps to Reproduce
1. told openclaw to clone the repo and run it
2. went to add agents
3. the error popped up

## Expected Behavior
i expected it to add a new agent

## Screenshots
![Image](https://github.com/user-attachments/assets/252c0edc-d817-4936-a9ed-f3fa657ba5e4)

## Environment
- Windows
- Brave browser
- Mission Control v 2. 0. 1

0xNyk:
Thanks for the report. `spawn openclaw ENOENT` usually means the `openclaw` executable is not installed or not on PATH for the Mission Control server process (common on Windows installs).

Please share:
1) output of `where openclaw` (PowerShell/CMD)
2) whether Mission Control was started from the same shell where `openclaw` is available
3) logs around `/api/spawn` request failure

Quick workaround: set `OPENCLAW_BIN` (or install OpenClaw and ensure PATH includes its install dir), then restart Mission Control.

0xNyk:
The `Spawn openclaw Enoent` error means the `openclaw` binary is not found on your system PATH. On Windows, this is common if OpenClaw was installed but the PATH was not updated.

**To fix:**
1. Make sure OpenClaw is installed (follow the OpenClaw installation guide)
2. Either add the OpenClaw binary directory to your system PATH, or set the `OPENCLAW_BIN` environment variable to the full path of the `openclaw` executable (e.g. `OPENCLAW_BIN=C:\Users\you\openclaw.exe` in your `.env` file)

The error message in Mission Control already provides this guidance: "Command not found: openclaw. Install it and ensure it is on PATH, or set OPENCLAW_BIN to an absolute executable path."

If OpenClaw is not installed, Mission Control can still run in standalone mode without a gateway — set `NEXT_PUBLIC_GATEWAY_OPTIONAL=true` in your `.env`.

0xNyk:
Heads-up — after today's merge of #648, the `isGatewayAvailable()` check was hardened to require an actual `openclaw.json` on disk before treating the OpenClaw subprocess path as live. That removes the spurious `spawn openclaw ENOENT` error in cases where the gateway DB has a seeded `primary` row but no OpenClaw binary is installed.

[...]

PR #648, opened 2026-05-05
feat(dispatch): direct multi-provider dispatch (Anthropic / OpenAI / local OpenAI-compatible)

nnnet:
## Summary

Adds a direct-dispatch path for tasks so MC no longer requires the OpenClaw gateway to reach LLMs. When the gateway isn't available — most commonly on Linux deployments where the macOS-native gateway can't be installed — MC now calls provider HTTP APIs itself.

### Supported direct providers

| Provider | Env variable | Notes |
|---|---|---|
| **Anthropic** | `ANTHROPIC_API_KEY` | Smart Opus/Sonnet/Haiku selection via `classifyDirectModel()` heuristics. |
| **OpenAI** | `OPENAI_API_KEY` | Models like `openai/gpt-4o-mini`, `openai/gpt-5-nano` selectable per-agent. |
| **Any OpenAI-compatible local** | `LOCAL_LLM_ENDPOINT`, optional `LOCAL_LLM_API_KEY` | LMStudio, Ollama, vLLM — anything serving `/v1/chat/completions`. |

Each agent's `agent_config.dispatchModel` overrides the heuristics (e.g. set `openai/gpt-4o-mini` to lock that agent to OpenAI).

### Hardened gateway-availability check

`isGatewayAvailable()` previously returned `true` whenever `~/.openclaw` was a non-empty path string — which is the **default** even when OpenClaw isn't installed. On Linux this caused `useDirectApi = !true && … = false`, so MC fell through to `runOpenClaw(...)` and crashed with `spawn openclaw ENOENT`.

The check now requires physical evidence:
- A real `openclaw.json` exists at `config.openclawConfigPath`, **or**
- A `gateways` row whose status is in `{online, healthy, ready}` (the seed `unknown` row planted by onboarding doesn't count).

Dropping the macOS-only constraint is the single biggest barrier to running MC in cloud / Linux dev environments. With this change, an operator can `git clone`, set `ANTHROPIC_API_KEY`, and dispatch tasks immediately — no gateway, no Mac, no extra processes.

## Walkthrough

`examples/MULTI-PROVIDER-DEMO.md` is a field-by-field walkthrough that builds a three-agent team using all three direct providers (Anthropic for the lead, OpenAI for the worker, local LMStudio for the reviewer) and runs an end-to-end task.

## Test plan

- Run on a Linux host with no OpenClaw installed; confirm `/agents` shows "No gateways installed" and dispatch still works.
- Set only `ANTHROPIC_API_KEY` and dispatch a task — logs show "Dispatching task via direct Claude API".
- Set only `OPENAI_API_KEY`, set agent `dispatchModel: openai/gpt-4o-mini`, dispatch — logs show OpenAI direct.
- Bring up LMStudio at `127.0.0.1:1234`, set `LOCAL_LLM_ENDPOINT=http://host.docker.internal:1234/v1`, dispatch with `dispatchModel: local/<your-model>` — logs show local direct.
- With OpenClaw running and a healthy `gateways` row, dispatch — `useDirectApi=false`, gateway path used as before. Backwards-compatible.

## Provenance

Squashes:
- `70328ad feat(dispatch): direct OpenAI + OpenAI-compatible local providers`
- `762ed51 fix(dispatch): make MC self-sufficient without OpenClaw gateway`
- `8def310 docs(deployment): document host-config projection, direct providers, host session modes`
- `bdd86ea docs(examples): add multi-provider team walkthrough`
- `c7627c7 docs(examples): expand multi-provider demo into a field-by-field walkthrough`

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -76,6 +76,100 @@ What `deploy:standalone` does:
 
 ## Production (Docker)
 
+Preferred operator flow (Make controls docker compose):
+
+```bash
+# 1) choose mode in .env
+#    MC_MODE=prod   # or dev
+#    OPENCLAW_ENABLED=1   # set 0 to run MC without OpenClaw stack
+
+# 2) run universal verbs
+make up
+make restart
+make down
+make status
+```
+
+### Mode-aware Make workflow (minimal commands)
+
+For day-to-day operations, see the [Daily Ops Cheatsheet](./ops-cheatsheet.md).
+
+Use `.env` + `.env.openclaw` as the single source of truth for mode/host/port/token values.
+
+- `MC_MODE=prod` → `docker-compose.yml`
+- `MC_MODE=dev` → `docker-compose-dev.yml`
+- `OPENCLAW_ENABLED=1` → `make <verb> all` includes OpenClaw
+- `OPENCLAW_ENABLED=0` → `make <verb> all` manages MC only
+
+Command grammar:
+
+```text
+make <verb> [all|mc|openclaw] [dev|prod]
+```
+
+- `all` is default scope.
+- `dev` / `prod` override `MC_MODE` for one command invocation.
+- Why no `--dev` / `--prod`: GNU Make consumes unknown `--xxx` tokens as Make options before Makefile goals are parsed, so mode overrides use positional tokens for deterministic behavior.
+- `make restart [scope]` is deterministic and always executes `make down [scope]` followed by `make up [scope]`.
+- With default `all` scope, `OPENCLAW_ENABLED=1` includes OpenClaw in both the down and up phases; `OPENCLAW_ENABLED=0` skips OpenClaw in both phases.
+
+Primary operator commands:
+
+| Workflow | Command |
+|---|---|
+| Start selected component(s) | `make up [all|mc|openclaw]` |
+| Restart selected component(s) | `make restart [all|mc|openclaw]` |
+| Stop selected component(s) | `make down [all|mc|openclaw]` |
+| Mode + endpoint health summary | `make status [all|mc|openclaw]` |
+| Refresh source/state only | `make update [all|mc|openclaw]` |
+| Force rebuild selected component(s) | `make rebuild [all|mc|openclaw]` |
+| Full maintenance (`update` + `rebuild` + `restart`) | `make upgrade [all|mc|openclaw]` |
+
+Mode override examples:
+
+```bash
+make restart dev
+make restart mc dev
+make status openclaw
+make upgrade prod
+```
+
+### `update` vs `upgrade`
+
+- `make update [scope]`
+  - Fast-forwards the current Mission Control branch from origin.
+  - For `scope=all`, if `OPENCLAW_ENABLED=1`, also refreshes OpenClaw source state.
+  - For `scope=openclaw`, refreshes OpenClaw source state regardless of `OPENCLAW_ENABLED`.
+  - Does **not** force an MC image rebuild and does **not** force restart.
+
+- `make upgrade [scope]`
+  - Runs update + rebuild + restart for selected scope.
+  - `scope=mc`: MC-only flow.
+  - `scope=openclaw`: OpenClaw update flow (`make openclaw-update`).
+  - `scope=all`: both flows; OpenClaw path runs when `OPENCLAW_ENABLED=1`.
+
+Minimum `.env` / `.env.openclaw` keys for this flow:
+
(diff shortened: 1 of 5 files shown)


0xbrainkid (changes requested):
Summary
- Blocking: direct dispatch availability/provider routing currently misclassifies common local-provider configurations and can enable direct mode even when no usable direct backend exists.

Strengths
- The gateway availability check is moving in the right direction by not trusting a default `~/.openclaw` path or unknown seeded gateway row.
- The UI defensive rendering for nested model config should prevent the React object-as-child crash.

Issues
- [BLOCKING] `isDirectDispatchAvailable()` returns true whenever `getLocalEndpoint()` returns a value, but `getLocalEndpoint()` defaults to `http://host.docker.internal:1234/v1`. That makes direct dispatch appear available on every no-gateway install, even with no API keys, no Claude CLI, and no local server. As a result `requeueStaleTasks()` skips stale/offline handling and `scoreAgentForTask()` treats offline agents as assignable, then dispatch can fail later (usually down the Anthropic path for default models). Please only treat local dispatch as available when it was explicitly configured and/or when checking availability for a selected local model.
- [BLOCKING] Local provider prefixes are stripped before provider selection. `classifyDirectModel()` returns only the last path segment for any `dispatchModel` (`local/qwen` -> `qwen`, `ollama/gpt-oss:20b` -> `gpt-oss:20b`), then `pickProvider()` no longer sees `local/`/`ollama/`; worse, an Ollama model starting with `gpt-` routes to OpenAI. Please preserve the provider prefix through `pickProvider()` and only strip it when constructing the downstream API request.

Questions
- Should default heuristic models (`claude-*`) require Anthropic key/Claude CLI specifically before direct mode can skip offline stale checks?

Verification
- `gh pr checks 648`: no checks reported.
- Attempted `pnpm typecheck`; it fails with existing dependency/type errors (e.g. missing `next-intl`/`node-pty`, Button `variant`/`size` props) in the worktree, so it was not a clean gate for this PR.

0xNyk (approved):
Approving — the gateway-free counterpart to #655 for #608.

**What it does:**
- Adds direct multi-provider dispatch (Anthropic / OpenAI / OpenAI-compatible local) so MC can dispatch tasks without OpenClaw being installed. Provider is chosen by the agent's `dispatchModel` prefix (`gpt-*` → OpenAI, `local/*`/`ollama/*`/`lmstudio/*` → `LOCAL_LLM_ENDPOINT`, anything else → Anthropic).
- For Anthropic: prefers the host-mounted Claude Code CLI (`/home/nextjs/.local/bin/claude` etc.) when available so the operator's existing login is used without exposing `ANTHROPIC_API_KEY` into the container; falls back to the API key path otherwise.
- Closes the gap on issue #609 too (OpenAI-compatible LLM as a provider — the local path covers Ollama / vLLM / liteLLM).

**Security review:**

1. **No new injection surface.** The new code paths still receive prompts that were already constructed by MC (task description, soul, etc.), and the existing injection-guard rules apply at the *task creation* boundary (preserved upstream). PR #648 isn't a place where new untrusted input enters the agent loop.
2. **`spawn('claude', args)` is safe.** Args are a fixed list (`--print`, `--output-format`, `--model`, `--append-system-prompt`); model and soul come from MC's DB; prompt is fed via stdin — never via shell concatenation. `CI: '1'` env, 180s timeout enforced. Not exploitable unless an attacker can already write to the DB.
3. **`fetch(`${endpoint}/chat/completions`)`** is a clean OpenAI REST call. Endpoint is operator-controlled via env (`LOCAL_LLM_ENDPOINT`), Authorization is properly bearer-headered. No SSRF.
4. **`isGatewayAvailable()` hardening is good** — it now requires either a real `openclaw.json` on disk OR a gateway row with status in (online/healthy/ready), instead of trusting the seeded `primary` row. That fixes the `spawn openclaw ENOENT` symptom in #588 as a side benefit.
5. **`requeueStaleTasks` skip in direct-API mode** is correct — agents have no heartbeat in direct-API mode by design, so the stale check would have failed every task after 5 cycles otherwise.

**Conflict note:** This PR overlaps `src/lib/task-dispatch.ts` with #655 which just landed. If the auto-merge produces a conflict, please rebase onto the new `main` — the changes are at different places in the file (you add `callDirectly`-family functions and route `useDirectApi` through them; #655 adds `reconcileDeferredTaskCompletions` and handles `agent.wait` timeouts as pending). Both should coexist cleanly.

Closes #608 (companion to #655) and #609. Merging.

PR #655, opened 2026-05-06
Fix async task dispatch reconciliation via programmatic gateway RPC

ChanchoNezz:
## Summary
- Replaces Mission Control task dispatch gateway calls with the programmatic websocket RPC path.
- Persists async dispatch metadata, including run/session identifiers.
- Treats `agent.wait` `status: "timeout"` as still pending instead of terminal.
- Reconciles terminal async completions into `review`.
- Recovers assistant response text from OpenClaw transcripts when terminal payloads do not include text.
- Adds focused regression coverage for gateway RPC, pending timeout handling, terminal promotion, missing run IDs, and transcript-backed result recovery.
## Validation
- `pnpm.cmd exec tsc --noEmit --pretty false`
- `pnpm.cmd vitest run src/lib/__tests__/gateway-runtime.test.ts src/lib/__tests__/openclaw-gateway.test.ts src/lib/__tests__/task-dispatch-reconciliation.test.ts src/lib/__tests__/task-status.test.ts`
- Live sandbox verification passed for `TASK-014`, including:
  - assigned -> in_progress
  - pending timeout remained pending/in_progress
  - terminal completion promoted to review
  - real assistant response text persisted as task resolution/comment

## Commit
`110e4c4 fix(tasks): reconcile async gateway dispatch completions`

PR #659, opened 2026-05-07
fix(recurring-tasks): include HH:MM in child title for sub-daily crons

### LINKED ISSUE #616: [Bug] Recurring tasks with sub-daily cron only spawn once per day — title-based dedup collision
## Bug Description

Recurring tasks with sub-daily cron schedules (e.g. `0 * * * *` hourly, `*/5 * * * *` every 5 min) only spawn **once per day**. All subsequent spawns are silently skipped.

## Root Cause

The duplicate-prevention check in `spawnRecurringTasks()` uses the child task **title** as the dedup key:

```typescript
const dateSuffix = formatDateSuffix()  // "Apr 24"
const childTitle = `${template.title} - ${dateSuffix}`  // "Memory Consolidation - Apr 24"

// Duplicate check:
const existing = db.prepare(
  `SELECT id FROM tasks WHERE title = ? AND workspace_id = ? AND project_id = ? LIMIT 1`
).get(childTitle, template.workspace_id, template.project_id)
if (existing) continue  // ← silently skips!
```

The `formatDateSuffix()` function only includes the **date** (`Mon DD`), not the time. So for an hourly cron like `0 * * * *`:

1. **02:00** — First spawn creates `"Memory Consolidation - Apr 24"` ✅
2. **03:00** — `isCronDue()` returns `true`, but the title `"Memory Consolidation - Apr 24"` already exists → **skipped** ❌
3. **04:00, 05:00, ...** — Same collision, all skipped ❌

## Impact

Any recurring task with a cron frequency higher than once-per-day is effectively limited to a single execution per day, regardless of the configured schedule.

## Proposed Fix

Extend `formatDateSuffix()` to include time (HH:MM) when the cron expression is sub-daily:

```diff
-function formatDateSuffix(): string {
+function formatDateSuffix(cronExpr?: string): string {
   const now = new Date()
   const months = [Jan, Feb, Mar, Apr, May, Jun, Jul, Aug, Sep, Oct, Nov, Dec]
-  return `${months[now.getMonth()]} ${String(now.getDate()).padStart(2, '0')}`
+  const datePart = `${months[now.getMonth()]} ${String(now.getDate()).padStart(2, '0')}`
+  if (cronExpr && isSubDailyCron(cronExpr)) {
+    const timePart = `${String(now.getHours()).padStart(2, '0')}:${String(now.getMinutes()).padStart(2, '0')}`
+    return `${datePart} ${timePart}`
+  }
+  return datePart
 }
```

With a helper `isSubDailyCron()`:

```typescript
function isSubDailyCron(cronExpr: string): boolean {
  const parts = cronExpr.trim().split(/\s+/)
  if (parts.length !== 5) return false
  const [minExpr, hourExpr] = parts
  if (hourExpr === '*') return true
  if (hourExpr.startsWith('*/')) {
    const step = parseInt(hourExpr.slice(2), 10)
    if (step > 0 && step < 24) return true
  }
  if (minExpr.startsWith('*/')) return true
  if (hourExpr.includes(',')) return true
  if (hourExpr.includes('-') && hourExpr !== '*') return true
  return false
}
```

And pass the cron expression at the call site:

```diff
-  const dateSuffix = formatDateSuffix()
+  const dateSuffix = formatDateSuffix(recurrence.cron_expr)
```

**Result:**
- Daily cron (`0 9 * * *`) → `"Task - Apr 24"` (unchanged)
- Hourly cron (`0 * * * *`) → `"Task - Apr 24 09:00"` (with time)
- Every 5 min (`*/5 * * * *`) → `"Task - Apr 24 09:05"` (with time)

## Testing

Tested locally with a `*/5 * * * *` recurring task. Three consecutive spawns produced unique titles without collisions:

| Time | Title | Status |
|------|-------|--------|
| 08:15 | `Test - Apr 24 08:15` | ✅ done |
| 08:20 | `Test - Apr 24 08:20` | ✅ done |
| 08:25 | `Test - Apr 24 08:25` | ✅ done |

## Suggested Improvement (future)

The current title-based dedup strategy is fragile. A more robust approach would be to deduplicate by `parent_task_id + scheduled_time` (e.g., a composite unique key on the parent-child relationship and the intended cron slot), making uniqueness independent of title formatting.

---

*This issue was identified and reported by an AI agent (OpenClaw). The bug analysis and proposed fix were tested locally before submission.*

0xNyk:
**Closes #616.**

## Problem

Recurring tasks with sub-daily cron schedules (e.g. \`0 * * * *\` hourly, \`*/5 * * * *\` every-5-min) only spawned once per calendar day. All subsequent spawns within the same day were silently skipped.

Root cause: \`spawnRecurringTasks\` derives child titles from \`formatDateSuffix()\` which only emitted day-level granularity (\`MMM DD\`). The duplicate-prevention guard at \`recurring-tasks.ts:75-79\` then matches on \`(title, workspace_id, project_id)\`, finds the previous spawn from earlier the same day, and returns early.

## Fix

1. New \`isSubDailyCron(cronExpr)\` helper returns true when the minute or hour field is anything other than a single concrete number (covers wildcards, ranges, lists, steps).
2. \`formatDateSuffix(now, subDaily)\` now takes an explicit \`now\` and a \`subDaily\` flag:

   | Cron shape | Suffix |
   |---|---|
   | Daily / weekly / monthly | \`Apr 24\` (unchanged) |
   | Sub-daily | \`Apr 24, 13:35\` |

   Same-day spawns at 13:00 and 14:00 now produce different titles, and the duplicate-guard correctly admits both.
3. Daily/weekly/monthly title format is preserved so existing operator workflows don't see a title-churn migration.

## Compatibility

- \`isCronDue()\` already prevents same-minute double-spawns via the \`lastSpawnedAtMs\` check, so this PR doesn't loosen the dedup — it just stops the dedup from firing on the *wrong* layer.
- The \`formatDateSuffix\` signature now takes optional arguments with sane defaults; existing call sites still work.

## Tests

13 new tests in \`src/lib/__tests__/recurring-tasks.test.ts\`:

- \`isSubDailyCron\` recognizes hourly / every-5-min / every-minute / ranged-minute crons; rejects daily/weekly; defensive on malformed.
- \`formatDateSuffix\` returns \`MMM DD\` (daily) and \`MMM DD, HH:MM\` (sub-daily); zero-pads day-of-month.
- **Regression**: two hourly spawns on the same day produce different titles.
- **Regression**: every-5-min spawns within the same hour produce different titles.
- Daily cron preserves historical \`MMM DD\` shape.

## Verification

\`\`\`bash
pnpm vitest run src/lib/__tests__/recurring-tasks.test.ts
# Test Files  1 passed (1)  Tests  13 passed (13)

pnpm tsc --noEmit
# (clean)
\`\`\`

## Note on the existing test failure

While running the full suite I noticed \`src/lib/__tests__/task-dispatch-reconciliation.test.ts\` has one pre-existing failing assertion on \`main\` that's unrelated to this PR (the test from PR #655's file expects \`callOpenClawGateway\` to be called once but gets zero calls). This PR does not touch \`task-dispatch.ts\` and the failure reproduces without my changes. Worth a separate look.

### DIFF
--- a/src/lib/__tests__/recurring-tasks.test.ts
+++ b/src/lib/__tests__/recurring-tasks.test.ts
@@ -1,0 +1,80 @@
+import { describe, it, expect } from 'vitest'
+import { isSubDailyCron, formatDateSuffix } from '@/lib/recurring-tasks'
+
+describe('isSubDailyCron', () => {
+  it('flags hourly crons (`0 * * * *`)', () => {
+    expect(isSubDailyCron('0 * * * *')).toBe(true)
+  })
+
+  it('flags every-5-min crons (`*/5 * * * *`)', () => {
+    expect(isSubDailyCron('*/5 * * * *')).toBe(true)
+  })
+
+  it('flags every-minute crons (`* * * * *`)', () => {
+    expect(isSubDailyCron('* * * * *')).toBe(true)
+  })
+
+  it('flags ranged-minute crons (`0,30 * * * *`)', () => {
+    expect(isSubDailyCron('0,30 * * * *')).toBe(true)
+  })
+
+  it('treats daily-at-09:00 as not sub-daily', () => {
+    expect(isSubDailyCron('0 9 * * *')).toBe(false)
+  })
+
+  it('treats weekly Mon-09:00 as not sub-daily', () => {
+    expect(isSubDailyCron('0 9 * * 1')).toBe(false)
+  })
+
+  it('returns false for malformed expressions (defensive)', () => {
+    expect(isSubDailyCron('not-a-cron')).toBe(false)
+    expect(isSubDailyCron('')).toBe(false)
+    expect(isSubDailyCron('0 9 *')).toBe(false)
+  })
+})
+
+describe('formatDateSuffix', () => {
+  // Pin a known date so test output is deterministic across machines/timezones
+  // we run the test in. We use a Date constructed via individual setters so
+  // local-time interpretation matches what the suffix uses.
+  const fixedDate = new Date(2026, 3, 24, 13, 35) // 2026-04-24 13:35 local
+
+  it('returns MMM DD for daily/weekly/monthly crons', () => {
+    expect(formatDateSuffix(fixedDate, false)).toBe('Apr 24')
+  })
+
+  it('returns MMM DD, HH:MM for sub-daily crons', () => {
+    expect(formatDateSuffix(fixedDate, true)).toBe('Apr 24, 13:35')
+  })
+
+  it('zero-pads day-of-month', () => {
+    const earlyMonth = new Date(2026, 0, 3, 9, 5) // Jan 3, 09:05
+    expect(formatDateSuffix(earlyMonth, false)).toBe('Jan 03')
+    expect(formatDateSuffix(earlyMonth, true)).toBe('Jan 03, 09:05')
+  })
+
+  it('regression #616: two hourly spawns on the same day produce different titles', () => {
+    const at13 = new Date(2026, 3, 24, 13, 0)
+    const at14 = new Date(2026, 3, 24, 14, 0)
+    const t13 = `Memory Consolidation - ${formatDateSuffix(at13, true)}`
+    const t14 = `Memory Consolidation - ${formatDateSuffix(at14, true)}`
+    // Before the fix both would have been "Memory Consolidation - Apr 24"
+    // and the second spawn would have been silently skipped.
+    expect(t13).not.toBe(t14)
+  })
+
+  it('regression #616: every-5-min spawns within an hour produce different titles', () => {
+    const at1300 = new Date(2026, 3, 24, 13, 0)
+    const at1305 = new Date(2026, 3, 24, 13, 5)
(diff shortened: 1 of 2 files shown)


0xNyk:
Self-merging via admin (GitHub blocks self-approval). Review notes for the audit trail:

- `isSubDailyCron` correctly detects wildcards / ranges / lists / steps in the minute or hour field. Defensive on malformed input (returns false rather than throwing).
- `formatDateSuffix` signature is backward-compatible (both new args optional, sane defaults).
- Daily/weekly/monthly title format preserved — no migration required for existing operator workflows that grew up on the `MMM DD` shape.
- `isCronDue()` already prevents same-minute double-spawns via `lastSpawnedAtMs` — this PR doesn't loosen the dedup, it stops it from firing on the wrong layer.
- 13 new tests including two explicit regression cases for the bug (hourly + every-5-min same-day).

Note: pre-existing failure in `task-dispatch-reconciliation.test.ts` on main is unrelated to this PR (file from #655) — flagged in the PR body for separate triage.

PR #669, opened 2026-05-09
fix: prevent @mention dropdown arrow-key selection reset

### LINKED ISSUE #661: [Bug] @mention autocomplete arrow-key selection resets to top of list
## Description

When typing `@` in a task comment to mention an agent, the autocomplete dropdown appears. Using arrow keys (Up/Down) to navigate the list works momentarily, but the selection immediately resets back to the first item. This makes it impossible to select agents below the first entry using the keyboard.

## Steps to Reproduce

1. Open a task comment box
2. Type `@` to trigger mention autocomplete
3. Press ArrowDown to move selection to the second item
4. Selection jumps back to the first item

## Expected Behavior

Arrow key navigation should cycle through the autocomplete list and maintain the selected index until Enter is pressed or the dropdown is dismissed.

## Code Location and Root Cause

*Analysis by Claude (AI assistant)*

**File:** `src/components/chat/chat-input.tsx` (~lines 172-185)

The `handleChange` handler unconditionally resets `mentionIndex` to 0 on every input event:

```typescript
const handleChange = (e: React.ChangeEvent<HTMLTextAreaElement>) => {
  const value = e.target.value
  setChatInput(value)
  const cursorPos = e.target.selectionStart
  const textBeforeCursor = value.slice(0, cursorPos)
  const atMatch = textBeforeCursor.match(/@(\w*)$/)
  if (atMatch) {
    setMentionFilter(atMatch[1])
    setShowMentions(true)
    setMentionIndex(0)  // ← always resets, even during arrow key nav
  } else {
    setShowMentions(false)
  }
}
```

When arrow keys trigger a re-render, `handleChange` fires again and resets `mentionIndex` to 0, overriding the navigation.

## Suggested Fix

Only reset `mentionIndex` when the filter text actually changes:

```typescript
const handleChange = (e: React.ChangeEvent<HTMLTextAreaElement>) => {
  const value = e.target.value
  setChatInput(value)
  const cursorPos = e.target.selectionStart
  const textBeforeCursor = value.slice(0, cursorPos)
  const atMatch = textBeforeCursor.match(/@(\w*)$/)
  if (atMatch) {
    const newFilter = atMatch[1]
    if (newFilter !== mentionFilter) {
      setMentionIndex(0)  // only reset when filter text changes
    }
    setMentionFilter(newFilter)
    setShowMentions(true)
  } else {
    setShowMentions(false)
  }
}
```

Additionally, guard against out-of-bounds index when the filtered list shrinks:

```typescript
useEffect(() => {
  if (mentionIndex >= filteredAgents.length) {
    setMentionIndex(Math.max(0, filteredAgents.length - 1))
  }
}, [filteredAgents.length, mentionIndex])
```

## Environment

- Mission Control: latest Docker image
- Browser: Chrome
- Multiple agents registered (6), so the dropdown has enough items to require navigation

tamish-max:
Fixes #661

The mention dropdown index was reset to 0 on every input change, making arrow-key navigation impossible. Now only resets when the filter text changes.

### DIFF
--- a/src/components/chat/chat-input.tsx
+++ b/src/components/chat/chat-input.tsx
@@ -144,9 +144,12 @@ export function ChatInput({ onSend, onAbort, disabled, agents = [], isGenerating
     const atMatch = textBeforeCursor.match(/@(\w*)$/)
 
     if (atMatch) {
-      setMentionFilter(atMatch[1])
+      const newFilter = atMatch[1]
+      if (newFilter !== mentionFilter) {
+        setMentionIndex(0)
+      }
+      setMentionFilter(newFilter)
       setShowMentions(true)
-      setMentionIndex(0)
     } else {
       setShowMentions(false)
     }


0xNyk:
Correct fix, and thank you for the patience — this has been open since May.

The old code reset `mentionIndex` to 0 on every keystroke that matched the @-pattern, so arrow-key movement in the dropdown was clobbered on the next render. Gating the reset on `newFilter !== mentionFilter` keeps the selection stable while the filter is unchanged, which is the actual intent.

Audited: UI state only, no new findings against the `main` baseline.

0xNyk:
One thing before this can land: the branch is behind `main`. Please rebase and it's good to go — the change itself needs nothing.

Branch protection requires `quality-gate` and CodeQL to pass, and fork workflow runs are held pending maintainer approval, so those need releasing too.

Issue #670, opened 2026-05-09
[bug] /api/security-audit crashes on SecretRef-form gateway.auth.token (TypeError: trim is not a function)


NePav:
## Summary

`/api/security-audit` crashes with `TypeError: ((intermediate value) ?? "").trim is not a function` whenever `openclaw.json` has `gateway.auth.token` (or `gateway.auth.password`) configured as a SecretRef object instead of a plaintext string.

This is the failure path triggered by the recommended secure setup: vault-backed gateway auth tokens (`{source:"file",provider:"vault",id:"/gateway-token"}`).

## Environment

- **Mission Control**: `2.0.1` (built from source on `~/repos/mission-control-app/`, native launchd service on macOS, not Docker)
- **OpenClaw gateway**: `2026.5.7 (eeef486)`
- **Reproduction**: 100% — any `openclaw.json` where `gateway.auth.mode = "token"` and `gateway.auth.token` is a SecretRef object

```jsonc
"gateway": {
  "auth": {
    "mode": "token",
    "token": {
      "source": "file",
      "provider": "vault",
      "id": "/gateway-token"
    }
  }
}
```

## Source

`src/lib/security-scan.ts:309-310`:

```ts
const tokenOk = gwAuth?.mode === 'token' && (gwAuth?.token ?? '').trim().length > 0
const passwordOk = gwAuth?.mode === 'password' && (gwAuth?.password ?? '').trim().length > 0
```

`gwAuth?.token` is the parsed JSON value from `openclaw.json`. When it's a SecretRef object, `(object ?? '')` returns the object (truthy/defined), and `.trim()` fails because objects don't have a `trim` method.

## Stack trace

```
TypeError: ((intermediate value) ?? "").trim is not a function
    at /Users/.../mission-control-app/.next/standalone/.next/server/chunks/src_lib_security-scan_ts_b78fd791._.js:1:4207
    at d (.../src_lib_security-scan_ts_b78fd791._.js:1:8904)
    at w (.../[root-of-the-server]__b4ace8c9._.js:1:16069)
    ...
    msg: "GET /api/security-audit error"
```

The error fires repeatedly because the security-audit endpoint is polled (UI badge / dashboard refresh).

## Suggested fix

Treat any non-string token/password value as "configured" for the purpose of the security-scan check (since presence of a SecretRef IS valid auth setup, and resolving the SecretRef requires the gateway runtime that security-scan doesn't have access to):

```ts
[...]

PR #674, opened 2026-05-12
fix(security-scan): handle SecretRef gateway auth credentials

NePav:
## Summary

Fixes a crash in `runSecurityScan` when `gateway.auth.token` or `gateway.auth.password` is a SecretRef object (e.g. `{ source: 'vault', ref: 'op://...' }`, `{ source: 'file', path: '/path/to/token' }`) rather than a plain string. The current code calls `.trim()` directly on these values, which throws:

```
TypeError: ((intermediate value) ?? "").trim is not a function
    at runSecurityScan (src/lib/security-scan.ts:312)
```

This makes both `GET /api/security-audit` and `POST /api/security-scan` return HTTP 500 for any install using SecretRef-based gateway auth — a common pattern with vault integrations on OpenClaw 2026.5+. The /debug page surfaces it as "Security scan error" in the logs, and the Settings panel's "Run security scan" button silently fails.

## Repro

`~/.openclaw/openclaw.json` snippet that triggers it:

```jsonc
{
  "gateway": {
    "auth": {
      "mode": "token",
      "token": { "source": "file", "path": "/Users/me/.openclaw/secrets/gateway-token" }
    }
  }
}
```

Click "Run security scan" in Settings → 500. Server logs show the TypeError above.

## Fix

Introduces a small `hasCredential()` helper in the `gateway_auth` check that treats:
- **string** → trim and check length (original behavior, no change)
- **non-null object** (SecretRef) → credential is configured; actual resolution happens at runtime by OpenClaw

If both branches fail (undefined / null), behavior is identical to before.

```ts
const hasCredential = (value: unknown): boolean => {
  if (typeof value === 'string') return value.trim().length > 0
  return value != null
}
const tokenOk = gwAuth?.mode === 'token' && hasCredential(gwAuth?.token)
const passwordOk = gwAuth?.mode === 'password' && hasCredential(gwAuth?.password)
```

## Test plan

- [x] `pnpm typecheck` clean
- [x] `pnpm vitest run src/lib/__tests__/security-scan` — 4/4 pass (existing tests still green; no behavior change for the string path they cover)
- [x] Manual: with SecretRef-typed `gateway.auth.token`, `GET /api/security-audit` now returns 200 with `gateway_auth: pass` instead of 500
- [x] Manual: with plain-string `gateway.auth.token`, scan still returns the same pass/fail as before
- [x] Manual: with `gateway.auth.token = ''`, scan still returns `gateway_auth: fail` as before
- [x] Manual: with no `gateway.auth` at all, scan still returns `gateway_auth: fail` as before

## Notes

- This does not validate that the SecretRef itself resolves to a non-empty string — that check would require live resolution and belongs in a separate runtime check. The intent of the `gateway_auth` security item is "is auth configured at all", which a SecretRef satisfies.
- I'm happy to add a dedicated unit test for the gateway_auth check covering string / SecretRef / undefined / empty string cases if preferred — let me know.

### DIFF
--- a/src/lib/security-scan.ts
+++ b/src/lib/security-scan.ts
@@ -309,8 +309,16 @@ function scanOpenClaw(): Category {
   } catch { /* skip */ }
 
   const gwAuth = ocConfig?.gateway?.auth
-  const tokenOk = gwAuth?.mode === 'token' && (gwAuth?.token ?? '').trim().length > 0
-  const passwordOk = gwAuth?.mode === 'password' && (gwAuth?.password ?? '').trim().length > 0
+  // gateway.auth.token / .password may be a plain string OR a SecretRef object
+  // (e.g. {source:"vault", ref:"op://..."} or {source:"file", path:"..."}).
+  // Calling .trim() on the object crashes. Treat any non-null value as
+  // "credential configured" — the resolved value is checked at runtime by OpenClaw.
+  const hasCredential = (value: unknown): boolean => {
+    if (typeof value === 'string') return value.trim().length > 0
+    return value != null
+  }
+  const tokenOk = gwAuth?.mode === 'token' && hasCredential(gwAuth?.token)
+  const passwordOk = gwAuth?.mode === 'password' && hasCredential(gwAuth?.password)
   const authOk = tokenOk || passwordOk
   checks.push({
     id: 'gateway_auth',


0xNyk:
Thanks @NePav — both the bug report (#670) and this fix were excellent: clear repro, correct root cause, and your `hasCredential()` typeguard is exactly the right shape (string → trim+length, non-null object → treat as configured since OpenClaw resolves the SecretRef at runtime).

We've landed this on `main` in commit `de01607`, using your `hasCredential()` approach verbatim and adding a regression test (`security-scan-gateway-auth.test.ts`) covering the string, empty-string, missing, and SecretRef (`{source:'file'}` / `{source:'vault'}`) forms so this can't regress.

Closing as superseded by the merged fix for #670. Thanks for catching this — it was a real 500 for anyone using vault-backed gateway auth. 🙏

PR #681, opened 2026-05-17
feat(api-client): add global 401/403/5xx interceptor + AuthExpiredListener

jun261930-tech:
## Summary

Fixes the UX bug where expired sessions silently fail — API calls return 401 but users see no feedback and stay on the current page.

### New files
- `src/lib/api-client.ts` — typed fetch wrapper with global 401/403/5xx interception
- `src/lib/__tests__/api-client.test.ts` — 8 vitest test cases covering all interception branches
- `src/components/auth-expired-listener.tsx` — Client Component that listens for `mc:auth-expired` and redirects to `/login?from=<current-path>`

### Modified files
- `src/app/layout.tsx` — mounts `<AuthExpiredListener />` at root layout
- `eslint.config.mjs` — adds `no-restricted-syntax` warn rule for bare fetch('/api/...) calls (3 quote variants)

## Verification

```
vitest: 8/8 PASS (2.5s)
typecheck: 0 errors
lint: 0 errors, 343 warnings (334 existing bare-fetch baselines)
```

## Migration path

See `PR-api-client.md` in workspace for 3-phase incremental migration strategy.

## Related
- PLAN.md Phase 1 deliverable
- Closes the auth-expired silent failure issue documented in SENIOR-CODE-REVIEW.md

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -18,6 +18,37 @@ const config = [
       'react-hooks/immutability': 'off',
     },
   },
+  // P1.5 — Discourage bare `fetch('/api/...')`.
+  // The team must migrate to `apiFetch<T>()` from `@/lib/api-client` so that
+  // 401 / 403 / 5xx / network failures are handled uniformly.
+  // Phase 1 (this PR): warn level, allow incremental migration.
+  // Phase 3 (final cleanup): upgrade to error and forbid merges that introduce
+  // new bare fetch('/api/...') sites.
+  // Selector rationale:
+  //   - covers single-quoted, double-quoted, and template-literal forms
+  //   - filters by /api prefix so cross-origin / external fetches stay untouched
+  //   - exempts api-client.ts itself (the one allowed implementer)
+  {
+    files: ['src/**/*.{ts,tsx,js,jsx}'],
+    ignores: ['src/lib/api-client.ts'],
+    rules: {
+      'no-restricted-syntax': [
+        'warn',
+        {
+          selector:
+            "CallExpression[callee.name='fetch'] > Literal[value=/^\\/api\\//]",
+          message:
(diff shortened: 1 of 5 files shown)


0xNyk:
Thank you for this whole series, @jun261930-tech — the `apiFetch` design is excellent: the typed `ApiError` codes, the 401→`/login` redirect with a redirect-loop guard, the `mc:auth-expired` event hook, and the eslint rule to enforce migration are exactly the right shape. We security-audited it and independently re-verified: same-origin `/api/` only, the `sameSite=strict` cookie can't leak cross-origin even with `credentials:'include'`, and the `/login?from=` param is inert (never read anywhere), so no open-redirect.

We've **landed the full series** on a consolidated branch:
- **Foundation** (this PR): `api-client.ts` + `AuthExpiredListener` + the eslint rule, applied verbatim (dropped only the non-functional `TemplateLiteral` esquery selector, which didn't match template-literal calls).
- **Migration** (#682–#697): all **188 `fetch('/api/...')` sites** across 21 components migrated to `apiFetch`.

Why reconstructed rather than merged directly: the 17 PRs were each branched independently and re-included the same foundation files, so they were mutually conflicting (several already showed `CONFLICTING`) and were cut from a pre-hardening base that predated the OpenCode runtime + other `main` changes. Merging them in sequence would have reverted post-PR work on several files. So we re-derived each file's migration on current `main`, preserving your exact approach — and in the process fixed a latent bug in the original chat migration where `conversation-list` computed `validSessions`/`validPrefs` but never used them (it kept mapping over the raw, unparsed response).

Verified: 0 bare `/api/` fetches remain, `tsc` clean, 969 unit tests pass, no new lint warnings. Closing as landed — your design carries through unchanged. This was a genuinely high-quality contribution; thank you. 🙏

PR #691, opened 2026-05-17
refactor(onboarding-wizard): migrate 10 bare fetch to apiFetch

jun261930-tech:
## Summary
Migrate 10 bare `fetch('/api/...')` calls in `src/components/onboarding/onboarding-wizard.tsx` to `apiFetch` for unified 401/403/5xx handling.

## Changes
- All 10 calls converted to `apiFetch<Response>(..., { raw: true })`
- Promise.allSettled parallel loading preserved
- Gateway health check preserved via `{ raw: true }`

## Why
Part of the P2.5 unified API client rollout. Replaces ad-hoc 401 handling with the centralized interceptor that emits `mc-auth-expired` and redirects to `/login`.

## Validation
- typecheck: 0 errors
- lint: 0 errors (file-level)
- No behavior change: error handling, status checks, body shape preserved

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -18,6 +18,37 @@ const config = [
       'react-hooks/immutability': 'off',
     },
   },
+  // P1.5 — Discourage bare `fetch('/api/...')`.
+  // The team must migrate to `apiFetch<T>()` from `@/lib/api-client` so that
+  // 401 / 403 / 5xx / network failures are handled uniformly.
+  // Phase 1 (this PR): warn level, allow incremental migration.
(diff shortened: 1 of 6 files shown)


0xNyk:
Landed as part of the consolidated apiFetch migration (see #681 for the full writeup). Your `onboarding-wizard` migration is included — all bare `fetch('/api/...')` sites in it now go through `apiFetch`. Reconstructed on current `main` rather than merged directly because the series was mutually conflicting on a stale base; your approach carries through unchanged. Thanks @jun261930-tech! 🙏

PR #692, opened 2026-05-17
refactor(agent-squad-panel-phase3): migrate 15 bare fetch to apiFetch

jun261930-tech:
## Summary
Migrate 15 bare fetch calls in `agent-squad-panel-phase3.tsx` to `apiFetch`.

## Changes
- 4 parallel-fetch calls use `apiFetch<Response>(..., { raw: true })` to preserve `res.ok` semantics
- `working_memory ?? ''` default added for optional API field

## Validation
- typecheck: 0 errors
- lint: 0 errors (file-level)

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -18,6 +18,37 @@ const config = [
       'react-hooks/immutability': 'off',
     },
   },
+  // P1.5 — Discourage bare `fetch('/api/...')`.
+  // The team must migrate to `apiFetch<T>()` from `@/lib/api-client` so that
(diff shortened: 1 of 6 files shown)


0xNyk:
Landed as part of the consolidated apiFetch migration (see #681 for the full writeup). Your `agent-squad-panel-phase3` migration is included — all bare `fetch('/api/...')` sites in it now go through `apiFetch`. Reconstructed on current `main` rather than merged directly because the series was mutually conflicting on a stale base; your approach carries through unchanged. Thanks @jun261930-tech! 🙏

PR #693, opened 2026-05-17
refactor([[...panel]]/page): migrate 13 bare fetch to apiFetch

jun261930-tech:
## Summary
Migrate 13 bare fetch calls in `src/app/[[...panel]]/page.tsx` to `apiFetch`.

## Changes
- `/api/auth/me` leverages apiFetch's default 401 → `/login` behavior
- All `OpenclawUpdate` payload fields use `?? ''` default to satisfy required-string types
- All variant-state branches (PUT/PATCH/POST) use `{ raw: true }` to keep `res.ok` semantics

## Validation
- typecheck: 0 errors
- lint: 0 errors (file-level)

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -18,6 +18,37 @@ const config = [
       'react-hooks/immutability': 'off',
     },
   },
+  // P1.5 — Discourage bare `fetch('/api/...')`.
+  // The team must migrate to `apiFetch<T>()` from `@/lib/api-client` so that
+  // 401 / 403 / 5xx / network failures are handled uniformly.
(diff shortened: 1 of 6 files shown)


0xNyk:
Landed as part of the consolidated apiFetch migration (see #681 for the full writeup). Your `[[...panel]]/page` migration is included — all bare `fetch('/api/...')` sites in it now go through `apiFetch`. Reconstructed on current `main` rather than merged directly because the series was mutually conflicting on a stale base; your approach carries through unchanged. Thanks @jun261930-tech! 🙏

PR #694, opened 2026-05-17
refactor(memory-browser-panel): migrate 12 bare fetch to apiFetch

jun261930-tech:
## Summary
Migrate 12 bare fetch calls in `memory-browser-panel.tsx` to `apiFetch`.

## Changes
- Memory list / detail / search / pin / archive / delete flows now use `apiFetch`
- `{ raw: true }` preserves status branching

## Validation
- typecheck: 0 errors
- lint: 0 errors (file-level)

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -18,6 +18,37 @@ const config = [
       'react-hooks/immutability': 'off',
     },
   },
+  // P1.5 — Discourage bare `fetch('/api/...')`.
+  // The team must migrate to `apiFetch<T>()` from `@/lib/api-client` so that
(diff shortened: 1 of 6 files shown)


0xNyk:
Landed as part of the consolidated apiFetch migration (see #681 for the full writeup). Your `memory-browser-panel` migration is included — all bare `fetch('/api/...')` sites in it now go through `apiFetch`. Reconstructed on current `main` rather than merged directly because the series was mutually conflicting on a stale base; your approach carries through unchanged. Thanks @jun261930-tech! 🙏

PR #695, opened 2026-05-17
refactor(runtime-setup-modal): migrate 12 bare fetch to apiFetch

jun261930-tech:
## Summary
Migrate 12 bare fetch calls across the runtime setup modal (OpenClaw / Hermes / Claude / Codex installers).

## Changes
- All 12 calls converted to `apiFetch<Response>(..., { raw: true })`
- Preserves `res.ok` / `res.status` / `data.X` handling for installer flows
- 401 now redirects to /login automatically across runtime setup

## Validation
- typecheck: 0 errors
- lint: 0 errors (file-level)

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -18,6 +18,37 @@ const config = [
       'react-hooks/immutability': 'off',
     },
   },
+  // P1.5 — Discourage bare `fetch('/api/...')`.
+  // The team must migrate to `apiFetch<T>()` from `@/lib/api-client` so that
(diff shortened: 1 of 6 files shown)


0xNyk:
Landed as part of the consolidated apiFetch migration (see #681 for the full writeup). Your `runtime-setup-modal` migration is included — all bare `fetch('/api/...')` sites in it now go through `apiFetch`. Reconstructed on current `main` rather than merged directly because the series was mutually conflicting on a stale base; your approach carries through unchanged. Thanks @jun261930-tech! 🙏

PR #703, opened 2026-05-19
fix(deps): extend pnpm.onlyBuiltDependencies for pnpm@11 compatibility

### LINKED ISSUE #702: Docker build fails with pnpm@latest: ERR_PNPM_IGNORED_BUILDS due to incomplete onlyBuiltDependencies allowlist
## Symptom

Running the documented Docker install path on a fresh checkout fails during `pnpm install` with `ERR_PNPM_IGNORED_BUILDS`. Native modules (notably `better-sqlite3`) are never compiled, so the postinstall verify step in the Dockerfile (`node -e "try{require('better-sqlite3')}..."`) exits non-zero and the build aborts.

```
ERR_PNPM_IGNORED_BUILDS  Ignored build scripts: @parcel/watcher@2.5.6, @swc/core@1.15.18,
better-sqlite3@12.6.2, esbuild@0.21.5, node-pty@1.1.0, sharp@0.34.5, unrs-resolver@1.11.1,
vue-demi@0.14.10.
```

## Reproduction

On a clean clone of `main`:

```bash
git clone https://github.com/builderz-labs/mission-control.git
cd mission-control
bash install.sh --docker
```

The build fails during the `pnpm install` layer.

## Root cause

Two things combine:

1. `Dockerfile` line 2 pins pnpm to whatever is current:
   ```dockerfile
   RUN corepack enable && corepack prepare pnpm@latest --activate
   ```
   This currently resolves to **pnpm v11.1.3**.

2. pnpm 11 changed the default: native build scripts are blocked unless the package is listed in `pnpm.onlyBuiltDependencies` in `package.json` (or approved via `pnpm approve-builds`).

The repo already has a `pnpm.onlyBuiltDependencies` field, but it only contains two entries:

```json
"pnpm": {
  "onlyBuiltDependencies": ["better-sqlite3", "node-pty"]
}
```

Five additional dependencies in the current lockfile legitimately need their build scripts to run: `@parcel/watcher`, `@swc/core`, `esbuild`, `sharp`, and `unrs-resolver`. (`vue-demi` also appears in the warning, but its postinstall is a no-op and can be safely left out.)

Because `better-sqlite3`'s build script is silently skipped, the hard failure surfaces later when the Dockerfile's postinstall verify step can't `require('better-sqlite3')`.

## Suggested fix

Extend `pnpm.onlyBuiltDependencies` in `package.json` to include the five missing packages:

```diff
   "pnpm": {
-    "onlyBuiltDependencies": ["better-sqlite3", "node-pty"]
+    "onlyBuiltDependencies": [
+      "@parcel/watcher",
+      "@swc/core",
+      "better-sqlite3",
+      "esbuild",
+      "node-pty",
+      "sharp",
+      "unrs-resolver"
+    ]
   }
```

Verified locally: with this change, `bash install.sh --docker` completes cleanly against pnpm 11.1.3 with no warnings.

Optional follow-up (not required to fix this bug): pin pnpm to a tested major in the Dockerfile (e.g. `corepack prepare pnpm@11 --activate`) so future major-version default changes don't silently break fresh installs again.

## Severity

**Blocks fresh Docker installs.** The `install.sh --docker` path is the documented setup flow in the README, so any new contributor or operator running it today against current pnpm hits this. Existing images built before pnpm 11 went out are unaffected.

## Notes

- Possibly related but distinct from closed #596 (OOM during build) and #323 (missing module).
- Error reproduces deterministically on a fresh checkout.
- A PR with this patch is incoming.

adm-jdlc:
Fixes #702.

## Summary

Extends `pnpm.onlyBuiltDependencies` in `package.json` from 2 to 7 entries so fresh Docker installs work cleanly against pnpm 11.

The current allowlist (`better-sqlite3`, `node-pty`) was correct for pnpm 10's permissive default but is incomplete now that pnpm 11 blocks unlisted native build scripts. Five additional packages in the lockfile legitimately need their install/postinstall to run:

| Package | Why it needs build scripts |
|---|---|
| `@parcel/watcher` | Native compile via `node scripts/build-from-source.js` |
| `@swc/core` | WASM/platform-binary fetch via `node postinstall.js` |
| `esbuild` | Platform-binary fetch via `node install.js` |
| `sharp` | Platform-binary fetch (or build from source fallback) |
| `unrs-resolver` | NAPI native via `napi-postinstall ... check` |

`vue-demi` appears in the warning but is intentionally omitted — its postinstall is a no-op try/catch shim with no side effects, so allowlisting it would add noise.

## Test plan

- [x] `bash install.sh --docker` from a fresh checkout against pnpm 11.1.3 — completes with no `ERR_PNPM_IGNORED_BUILDS` warnings.
- [x] Container starts healthy: `GET /api/status?action=health` returns `{"status":"healthy"}`.
- [x] `/setup` flow + admin creation work end-to-end.
- [x] `pnpm install` (host-side, without Docker) also clean.

## Notes

- This is the minimal fix. A reasonable follow-up — not included here to keep the PR scoped — is pinning the Dockerfile to a tested pnpm major (`corepack prepare pnpm@11 --activate`) so the next major bump doesn't silently break installs again.
- See #702 for full root cause + repro details.

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,8 +95,13 @@
   },
   "pnpm": {
     "onlyBuiltDependencies": [
+      "@parcel/watcher",
+      "@swc/core",
       "better-sqlite3",
-      "node-pty"
+      "esbuild",
+      "node-pty",
+      "sharp",
+      "unrs-resolver"
     ]
   }
 }


adm-jdlc:
**Update — found a more nuanced root cause while testing locally.**

After landing this fix in my workspace, I ran `pnpm add` and got:

```
[WARN] The "pnpm" field in package.json is no longer read by pnpm.
The following keys were ignored: "pnpm.onlyBuiltDependencies".
See https://pnpm.io/settings for the new home of each setting.
```

So as of pnpm 11+, `pnpm.onlyBuiltDependencies` in `package.json` is **deprecated and ignored** — settings moved to `pnpm-workspace.yaml` under `allowBuilds:` (a map of package matchers to `true`/`false`).

Two takeaways:

1. **This PR's fix still helps users on pnpm 10.x** (which still reads the package.json field), so it's not wrong — but it's a transitional fix.

2. **For pnpm 11+, the canonical fix is** to add a `pnpm-workspace.yaml` like:

   ```yaml
   allowBuilds:
     '@parcel/watcher': true
     '@swc/core': true
     better-sqlite3: true
     esbuild: true
     node-pty: true
     sharp: true
     unrs-resolver: true
     vue-demi: false
   ```

Notably, when pnpm 11+ runs `pnpm add` with a missing/incomplete config, it auto-generates `pnpm-workspace.yaml` with placeholder values reading literally `set this to true or false` (not `true`/`false`), which means the warning persists silently until an operator notices.

Happy to amend this PR to add `pnpm-workspace.yaml` instead of (or alongside) the `package.json` change — or split into a follow-up PR if you'd prefer to keep this one minimal. Let me know which you'd like.

hombarsourcingpartner-git:
Thanks

its work

add this line too in docker file

COPY pnpm-workspace.yaml* ./

0xNyk:
Thanks for this, @adm-jdlc — your root-cause analysis in #702 was spot-on, including the correct call to exclude `vue-demi` (its postinstall is a no-op).

This has now landed on `main` in commit `9d2c5d4` (`fix(build): pin pnpm to v10 and complete onlyBuiltDependencies allowlist`), which uses the **exact same 7-package allowlist** you proposed here (`@parcel/watcher`, `@swc/core`, `better-sqlite3`, `esbuild`, `node-pty`, `sharp`, `unrs-resolver`) and additionally pins `pnpm@10.29.3` via both `packageManager` and the Dockerfile to match CI and keep builds deterministic.

Closing as superseded by #702. Really appreciate the thorough writeup. 🙏

PR #709, opened 2026-05-21
Run ANALYZE in auto_cleanup + make claude-session scan interval configurable

yojen7:
## Problem
After a SQLite maintenance/recover pass, stale planner statistics can make the query planner pick a poor index, spiking CPU on the tasks read path. Separately, the claude-session scan interval (TICK_MS) is hardcoded, so operators must fork to tune it.

## Change
- Add `ANALYZE` to the daily `auto_cleanup` scheduled task so planner stats stay fresh.
- Read the claude-session scan interval from `MC_CLAUDE_SCAN_INTERVAL_MS` (falling back to the current default) so it is tunable without a fork.

Low-risk maintenance/configurability hygiene; no schema or API changes.

### DIFF
--- a/src/lib/scheduler.ts
+++ b/src/lib/scheduler.ts
@@ -51,6 +51,14 @@ function getSettingNumber(key: string, defaultValue: number): number {
   }
 }
 
+function getEnvNumber(key: string, defaultValue: number): number {
+  const raw = process.env[key]
+  if (!raw) return defaultValue
+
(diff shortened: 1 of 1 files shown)


0xNyk:
Landed on `main` (security-audited + verified clean). ANALYZE now runs in auto_cleanup and the Claude scan interval is configurable via `MC_CLAUDE_SCAN_INTERVAL_MS`. Thanks @yojen7! 🙏

PR #710, opened 2026-05-21
Make session file size limit configurable via MC_MAX_SESSION_FILE_BYTES

yojen7:
## Problem
The maximum Claude session file size is a hardcoded constant; operators with larger sessions must fork to raise it.

## Change
Read the cap from `MC_MAX_SESSION_FILE_BYTES`, defaulting to the existing value. Pure configurability, no behavior change when unset.


0xNyk:
Landed on `main` (security-audited + verified clean). Session file size limit is now configurable via `MC_MAX_SESSION_FILE_BYTES` (default 50MB). Thanks @yojen7! 🙏

PR #668, opened 2026-05-08
Harden workspace-scoped route access

niko4244:
Hardens workspace-scoped API route access by removing silent workspace fallback, enforcing fail-closed workspace context, restricting agent-scoped API keys to their own heartbeat/memory and assigned task surfaces, and adding focused regression coverage for workspace-scope and same-workspace overreach cases.

Scope:
- Clean one-commit cherry-pick onto builderz-labs/main
- Commit: 1e09f7e Harden workspace-scoped route access
- Diff: 17 files changed
- Old oversized PR #667 should be closed and replaced by this PR

Security changes:
- Missing workspace context now fails closed on reviewed route paths
- Removed fallback to workspace 1 from reviewed routes/webhook bookkeeping
- Agent-scoped API keys are restricted to their own heartbeat and memory routes
- Agent-scoped API keys are restricted to their own assigned task/comment/queue surfaces
- Webhook CRUD uses workspace-scoped access

Validation note:
This branch is clean and isolated, but full repo validation currently fails on builderz-labs/main before this patch. The same failures are present on untouched upstream main and are unrelated to this PR.

Pre-existing upstream validation failures:
- pnpm typecheck: FAIL due to .next/dev/types/validator.ts missing-module errors and existing system-monitor-panel.tsx type issues
- pnpm test: FAIL in existing OpenCode/config/path/reconciliation test suites
- pnpm lint: FAIL in existing unrelated panel files
- pnpm build: FAIL during type checking on the same pre-existing validator/type issues

Patch-specific finding:
- No validation failure was introduced by this workspace-scope hardening patch.

PR #711, opened 2026-05-21
Improve GET /api/tasks read-path performance; make deferred reconciliation opt-in

yojen7:
## Problem
`GET /api/tasks` uses a correlated `COUNT` subquery for comment counts, which is fragile under stale planner stats and can pick a poor index. It also runs `reconcileDeferredTaskCompletions` on every call, adding write work to a hot read path.

## Change
- Rewrite the correlated comment-count subquery as a `LEFT JOIN comments ... GROUP BY t.id` (more robust to stale stats; fewer planner pitfalls).
- Gate `reconcileDeferredTaskCompletions` so it does not run on every `GET /api/tasks` (it is already scheduled separately).

Read-path performance + robustness; behavior-preserving for the returned payload.

### DIFF
--- a/src/app/api/tasks/route.ts
+++ b/src/app/api/tasks/route.ts
@@ -80,19 +80,24 @@ export async function GET(request: NextRequest) {
     const limit = Math.min(parseInt(searchParams.get('limit') || '50'), 200);
     const offset = parseInt(searchParams.get('offset') || '0');
 
-    try {
-      await reconcileDeferredTaskCompletions({ workspaceId, limit: 5 })
-    } catch (err) {
-      logger.warn({ err }, 'Deferred task reconciliation failed during task list read')
+    const shouldReconcileDeferred = ['1', 'true', 'yes'].includes((searchParams.get('reconcile_deferred') || '').toLowerCase());
+    if (shouldReconcileDeferred) {
+      try {
+        await reconcileDeferredTaskCompletions({ workspaceId, limit: 5 })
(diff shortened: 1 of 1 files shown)


0xNyk:
Thanks @yojen7 — the read-path-performance concern was real and your instinct (don't run a write/reconciliation on every list read) was correct. Reviewing against current `main`, though, this is largely already resolved:

1. **Deferred reconciliation on read** — the `reconcileDeferredTaskCompletions({ workspaceId, limit: 5 })` call this PR makes opt-in no longer exists in the codebase at all (it was removed from the `GET /api/tasks` path). `grep` for it returns zero hits, so there's no read-path write left to gate.

2. **comment_count subquery** — the correlated `(SELECT COUNT(*) FROM comments ...)` is backed by `idx_comments_task_id` (schema.sql:119), so for the 50-row default page it's indexed point lookups. The `LEFT JOIN comments + GROUP BY` rewrite is an equivalent plan but adds GROUP-BY fragility alongside `SELECT t.*`, with no clear win given the index.

Since the diff also doesn't apply (it's cut from a base before the workspace-scope hardening that just landed via #668), I'm closing this rather than force a marginal rewrite. If you still see slow `GET /api/tasks` on current `main` with a populated DB, please reopen with an `EXPLAIN QUERY PLAN` and row counts and we'll dig into the specific bottleneck. Appreciate the contribution and the performance focus. 🙏

PR #722, opened 2026-05-30
fix(docker): expand onlyBuiltDependencies for pnpm 11+

### LINKED ISSUE #702: Docker build fails with pnpm@latest: ERR_PNPM_IGNORED_BUILDS due to incomplete onlyBuiltDependencies allowlist

rsnobles:
Fixes #702

### Changes
- Expanded `pnpm.onlyBuiltDependencies` in `package.json` to include all native modules required by pnpm 11+.
- Pinned `Dockerfile` to `pnpm@11` (optional hardening).

This resolves `ERR_PNPM_IGNORED_BUILDS` during fresh Docker installs.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 FROM node:22.22.0-slim AS base
-RUN corepack enable && corepack prepare pnpm@latest --activate
+RUN corepack enable && corepack prepare pnpm@11 --activate
 WORKDIR /app
 
 FROM base AS deps
--- a/package.json
+++ b/package.json
@@ -95,8 +95,13 @@
   },
   "pnpm": {
     "onlyBuiltDependencies": [
-      "better-sqlite3",
-      "node-pty"
+    "@parcel/watcher",
+    "@swc/core",
+    "better-sqlite3",
+    "esbuild",
+    "node-pty",
+    "sharp",
+    "unrs-resolver"
     ]
   }
 }


0xNyk:
Thanks @rsnobles. The `onlyBuiltDependencies` expansion here matches what we needed, and it's now on `main` in commit `9d2c5d4` with the identical 7-package allowlist.

One note for the record: we pinned the Dockerfile to **`pnpm@10`** rather than `pnpm@11`, because pnpm 11 is what turns `ERR_PNPM_IGNORED_BUILDS` into a hard failure in the first place — pinning to 10 (matching CI and the committed `lockfileVersion 9.0`) is what keeps fresh Docker builds green. So the `package.json` half of this PR is already covered, and the Dockerfile half we intentionally took the other direction.

Closing as superseded by #702. Thanks for pitching in. 🙏

PR #731, opened 2026-06-01
fix(god-mode): expose stale loop status

JayRMorris:
## Summary
- add repo-owned God Mode status reader with MISSION_CONTROL_ARK_HOME test seam
- expose loop status freshness fields at top level and inside rich_loop_status
- ignore generated runtime/proof artifact directories from Noah UI improvement loops

## Proof
- pnpm install --frozen-lockfile (minimal worktree dependency install against origin/main lockfile)
- pnpm exec vitest run src/lib/__tests__/god-mode.test.ts  # 2 passed
- pnpm typecheck  # passed

## Notes
- The older local claude/god-mode-agi-stack-parity-superiority branch could not be pushed directly because its pack exceeded GitHub's 2 GiB limit from historical artifact commits, so this PR is the minimal clean branch from origin/main.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,17 @@ playwright-report/
 /e2e-debug-*.png
 /e2e-channels-*.png
 
+# Generated runtime/proof artifacts
+/artifacts/audits/
+/artifacts/branding-proof-*/
+/artifacts/hrm-synthetic-reasoning-*/
+/artifacts/noah-authenticated-ui-review-*/
+/artifacts/noah-bleeding-edge-readiness-*/
+/artifacts/noah-feature-probe-*/
+/artifacts/noah-hermes-readiness-*/
+/artifacts/noah-interpretability-readiness-*/
+/artifacts/noah-ui-improvement-loop-*/
+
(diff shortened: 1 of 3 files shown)


0xNyk:
Thanks @JayRMorris — reviewed in depth. The module is self-contained and clean (fs-only under ~/.ark, no exec/network), and the staleness logic itself is sound. The blocker is that it lands as dead code: nothing in this PR wires `god-mode.ts` to any route, panel, or dispatch path, so mainline would be carrying a 527-line personal subsystem with no caller. Also the `.gitignore` additions (`/artifacts/noah-*`) are specific to your local loops.

If you want this upstream: wire it to an API route + a visible surface (even a diagnostics panel entry), drop the personal gitignore entries, and trim the manifesto framing to code comments. Alternatively it may be happiest living in your fork. Leaving open for now — your call which direction to take.

0xNyk:
Closing for now per the earlier review: mainline can't carry an unwired 527-line module, and this looks happiest living in your fork until it has a caller. If you wire it to a route/panel and trim the personal artifacts, reopen or resubmit and I'll review promptly.

PR #747, opened 2026-06-26
Fix: restrict gateway bearer token exposure to operator+ roles

LHMisme420:
Fixes the issue reported in the security advisory: GET-style POST /api/gateways/connect
returned the real gateway bearer token to any authenticated viewer-role user, not just
operator/admin.

Root cause: the endpoint correctly stays open to `viewer` role (per the existing comment
about not breaking the startup fallback flow), but the response unconditionally included
the secret `token` regardless of the caller's role.

Fix: gate inclusion of the real token value on `auth.user.role >= operator`. Viewer-role
callers still get `ws_url` and connection metadata, just not the raw credential.

Open question for maintainer: please confirm the "startup fallback" flow mentioned in the
original comment authenticates at operator role or above — if it currently relies on a
viewer-level identity to retrieve the actual token, this PR would need an adjustment for
that specific internal path.

### DIFF
--- a/src/app/api/gateways/connect/route.ts
+++ b/src/app/api/gateways/connect/route.ts
@@ -1,5 +1,5 @@
 import { NextRequest, NextResponse } from 'next/server'
-import { requireRole } from '@/lib/auth'
+import { requireRole, ROLE_LEVELS } from '@/lib/auth'
 import { getDatabase } from '@/lib/db'
 import { buildGatewayWebSocketUrl } from '@/lib/gateway-url'
 import { getDetectedGatewayToken } from '@/lib/gateway-runtime'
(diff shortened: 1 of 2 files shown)


LHMisme420:
For context on how this was reported: I attempted to send this to
security@builderz.dev per SECURITY.md, but the email bounced. There's
no GitHub Security Advisory option enabled on this repo either, so this
PR is effectively serving as the disclosure record — happy to move this
to a private advisory if you enable one.

## Vulnerability details

POST /api/gateways/connect required only `viewer` role (the lowest tier)
to call, and returned the actual gateway bearer token in the JSON response
regardless of caller role. Any authenticated viewer-tier user could
extract the full credential and use it to connect directly to the
OpenClaw gateway with operator/admin-level capability — a straight
viewer-to-full-control privilege escalation.

CWE-862 (Missing Authorization) / CWE-269 (Improper Privilege Management)

Before merge — can you confirm the "startup fallback" flow referenced in
the original comment authenticates at operator role or above? If it relies
on a viewer-level identity to fetch the real token, this PR's gate would
break that path and need an adjustment.

0xNyk:
Reviewed — this genuinely narrows access, and the answer to your open question: the startup fallback in `src/app/[[...panel]]/page.tsx:226` runs with the caller's own session. For operator/admin nothing changes; for viewer-role sessions the response now degrades to `token: '', token_set: false`, which is the exact same signal the frontend already handles for 'no token configured' — so the flow degrades cleanly rather than breaking. That's the intended effect of the fix.

Two nits I'll fix in a follow-up rather than blocking: both files lose their trailing newline, and `/api/index` documents this route as `auth: 'operator'` while the handler gates at `viewer` — worth reconciling now that the token is role-gated. CI is approved; merging once green. Thanks for a careful fix.

PR #751, opened 2026-06-29
v2 model catalog overhaul + chat status filtering + dependency hardening

Obrais-cloud:
## Summary
Rebased cleanly onto current `main` (ahead 1 / behind 0, no conflicts). Verified locally: **lint, typecheck, test (717 passing), and build all green**.

### Model catalog (`src/lib/models.ts`)
- Replace `MODEL_CATALOG` with the current provider set: Anthropic 4.x, OpenAI gpt-5.x, Google, DeepSeek, xAI, MiniMax, NVIDIA, and blockrun smart-router aliases (`auto`/`eco`/`premium`).
- Extend `ModelConfig` with optional `reasoning` / `vision` flags and add `getModelsByProvider()`.
- `costPer1k` values reflect input $/MTok (e.g. Haiku 4.5 = 1.0, Sonnet 4.6 = 3.0, Opus = 5.0).

### Chat (`src/components/chat/message-list.tsx`)
- Hide transient status messages (`offline`/`delivery_failed`, `accepted`/`processing`) once a later text/tool_call or terminal status supersedes them.

### Tests
- `models.test.ts`: haiku cost assertion `< 1` → `<= 1` (matches Haiku 4.5 input price of $1/MTok).

### Dependencies (`package.json` / `pnpm-lock.yaml`)
- Add `pnpm.overrides` for `shell-quote`, `rollup`, and `minimatch` (removes ReDoS / path-traversal advisories; preserves upstream's existing pnpm config).
- Bump `vitest` and `@vitest/coverage-v8` to v3 (resolves the Vitest UI critical advisory). Lockfile regenerated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,11 +69,11 @@
     "@types/react-dom": "^19.0.3",
     "@types/ws": "^8.18.1",
     "@vitejs/plugin-react": "^4.3.4",
-    "@vitest/coverage-v8": "^2.1.9",
+    "@vitest/coverage-v8": "^3",
     "jsdom": "^26.0.0",
     "pino-pretty": "^13.1.3",
     "vite-tsconfig-paths": "^5.1.4",
-    "vitest": "^2.1.5"
+    "vitest": "^3"
   },
   "engines": {
     "node": ">=22"
@@ -103,6 +103,12 @@
       "unrs-resolver"
     ],
     "minimumReleaseAge": 10080,
-    "minimumReleaseAgeExclude": []
+    "minimumReleaseAgeExclude": [],
+    "overrides": {
+      "shell-quote": "^1.8.4",
+      "rollup": "^4.59.0",
+      "minimatch@<3.1.4": "3.1.4",
+      "minimatch@>=9.0.0 <9.0.7": "9.0.7"
+    }
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -4,6 +4,12 @@ settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
+overrides:
+  shell-quote: ^1.8.4
+  rollup: ^4.59.0
+  minimatch@<3.1.4: 3.1.4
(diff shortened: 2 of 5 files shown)


0xNyk:
Thanks @Obrais-cloud — the dependency-hardening part of this (pnpm overrides pinning shell-quote / rollup / minimatch above known-CVE ranges) is exactly right, and I'm going to land those pins on main directly this week as part of a security-deps pass (with credit) so they become the baseline. For the rest, requesting changes before this can merge:

1. **Model ID format**: the catalog renames `claude-haiku-4-5` → `claude-haiku-4.5` (dash→dot), but `classifyDirectModel()` in `src/lib/task-dispatch.ts` still string-matches the dashed form and this PR doesn't touch it — dot-form entries would silently fail direct-dispatch classification. Whichever format is right for the dispatch path, catalog + classifier must move together (ideally classifier derives from the catalog so there's one source of truth).
2. **Pricing changes need sources**: haiku 0.25→1.0 and opus 15→5 are big swings — please link the provider price pages in the PR description. Also note the field is named `costPer1k` but holds per-1M values; if we're overhauling the catalog anyway, renaming to `costPerMTok` would fix a 1000× footgun.
3. **Scope**: drop the vitest ^2→^3 major from this PR — majors now arrive as individual dependabot PRs (#755) and it conflicts with that pipeline.
4. Needs a rebase + green CI run once the above is addressed.

The chat status filtering piece looks fine. Keeping open.

0xNyk:
Superseding this via #756 (your pnpm overrides — landed, credited) and the new catalog PR (costPerMTok rename, catalog-derived classification, doc-verified prices — credited in the commit). The dot-format ID rename couldn't land as-is since `classifyDirectModel` string-matches the dashed IDs; the new PR makes the catalog the single source of truth so that class of drift can't recur. The chat status-filtering piece from this PR is still welcome as its own small PR if you'd like to resubmit it — that part was clean. Thanks again for pushing on the catalog staleness!

PR #773, opened 2026-07-11
fix(token-usage): correct INSERT columns so dispatch usage persists

ForaDoPadraoYT:
Problem: the 3 INSERT INTO token_usage in src/lib/task-dispatch.ts write to columns that do not exist (total_tokens, cost). Every insert throws and is swallowed by the non-fatal catch, so token_usage never records dispatch usage and the Cost Tracker reads zero for all providers. Real schema: model, session_id, input_tokens, output_tokens, created_at, workspace_id, task_id, cost_usd, agent_name.

Fix: align the 3 INSERTs with the real columns (drop total_tokens, rename cost -> cost_usd). Minimal, no behavior change beyond the inserts succeeding.

Verified: token_usage had 0 rows before; after the fix a real dispatch run persists a row.

### DIFF
--- a/src/lib/task-dispatch.ts
+++ b/src/lib/task-dispatch.ts
@@ -748,14 +748,13 @@ async function callClaudeDirectly(
       const db = getDatabase()
       const now = Math.floor(Date.now() / 1000)
       db.prepare(`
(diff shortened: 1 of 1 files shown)


0xNyk:
Superseded by #782, which includes the same token_usage schema correction, centralizes the insert to prevent future drift, adds regression coverage, and has passed the full quality gate. Thank you for identifying the underlying bug.

PR #778, opened 2026-07-12
fix: explain stale auto-backup alerts

ForaDoPadraoYT:
## What changed

- Make the backup security check read the `general.auto_backup` setting.
- Report when the scheduler is not registered, waiting for its first run, or when the last scheduled run failed.
- Keep the existing recent-backup age thresholds.
- Add focused tests for the stale-backup alert cases.

## Why

The previous warning always suggested enabling `auto_backup` when a backup was older than 24 hours, even when the setting was already enabled. This made an old backup look like a configuration problem instead of distinguishing configuration, scheduler, and execution state.

## Validation

- Focused backup alert tests passed.
- Typecheck passed.
- Production build passed.
- Lint passed with existing warnings and no errors.
- The full test suite still has an unrelated `gnap-sync.test.ts` environment failure because Git has no configured `user.name`/`user.email` in this WSL environment.

### DIFF
--- a/src/lib/__tests__/security-scan-backup.test.ts
+++ b/src/lib/__tests__/security-scan-backup.test.ts
@@ -1,0 +1,51 @@
+import { describe, expect, it } from 'vitest'
+import { buildBackupCheck } from '@/lib/security-scan'
+
+const schedulerRunning = {
+  enabled: true,
+  schedulerRegistered: true,
+  schedulerLastRun: Date.now() - 86_400_000,
+}
+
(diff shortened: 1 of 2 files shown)


0xNyk:
Thank you — the core observation here was valid. I adapted it on current main in #789 and credited this PR there. The replacement also avoids suggesting unauthenticated curl for the admin-only backup endpoint, preserves an unknown state when settings cannot be read, and keeps raw scheduler errors out of scan output. Closing this stale branch in favor of the hardened replacement.

Issue #194, opened 2026-03-04
WebSocket connection fails with Tailscale Serve and Cloudflare Tunnel


Coyote-Den:
Environment:

• Server: Oracle Cloud Free Tier (ARM)
• OS: Linux 6.17.0-1007-oracle (arm64)
• Mission Control: Latest from builderz-labs/mission-control
• OpenClaw Gateway: Running on same server, binds to 127.0.0.1:18789

Description:

When configuring Mission Control to connect to an OpenClaw gateway hosted on the same machine but accessed remotely via Tailscale or Cloudflare tunnel, the WebSocket connection consistently fails.

───

Steps to Reproduce:

1. Run OpenClaw gateway on server (bind: loopback)
2. Expose gateway via Tailscale Serve: tailscale serve http://127.0.0.1:18789
3. Or expose via Cloudflare Tunnel
4. Configure Mission Control gateway with Tailscale/Cloudflare URL
5. Attempt to connect - WebSocket fails with errors

───

Expected Behavior:

Mission Control should connect to OpenClaw gateway via remote URL (Tailscale or Cloudflare) and establish WebSocket connection for real-time updates.

───

Actual Behavior:

1. Tailscale Serve:
  • Gateway probe works (HTTP)
  • WebSocket fails: browser tries wss://cb-vcn.tail47c878.ts.net:18789/
  • Connection refused/closed immediately
  • Error: WebSocket error: ... readyState: 3 (CLOSED)
2. Cloudflare Tunnel:
  • Gateway probe fails with HTTP 530 or 400 errors
  • WebSocket cannot establish
  • Cloudflare blocks WebSocket upgrade requests
3. Client ID Error:
  • Default client ID control-ui rejected by gateway
  • Error: invalid connect params: at /client/id: must be equal to constant
  • Required fix: Use web instead of control-ui

───

Errors Observed:

WebSocket error: 
url: "wss://cb-vcn.tail47c878.ts.net:18789/"
readyState: 3

Gateway error: invalid connect params: at /client/id: must be equal to constant; at /client/id: must match a schema in anyOf

───

Workaround Attempted:

1. Changed default client ID from control-ui to web - partially helps
2. Tried different ports (443 vs 18789) - WebSocket still fails
3. Tried Cloudflare tunnel for gateway - 530/400 errors

───

Files Affected:

• src/lib/websocket.ts - WebSocket client (line 19: DEFAULT_GATEWAY_CLIENT_ID)
[...]

PR #832, opened 2026-07-16
fix(gateway): keep tunnel reconnects on stable websocket URL

leovinci-sd:
## Summary\n- Fix reconnect handling after a successful gateway WebSocket handshake drops later\n- Preserve the last known-good tunnel URL instead of misclassifying established disconnects as initial handshake failures\n- Prevent unintended fallback from a working root URL to /gateway-ws or /gw during normal tunnel reconnects\n\n## Why\nFollow-up for #194: users reported tunnel connections establishing and then disconnecting later. The close handler cleared handshake state before checking it, so every close looked like a failed initial handshake.\n\n## Testing\n- pnpm typecheck\n- pnpm exec vitest run src/lib/__tests__/gateway-url.test.ts\n\nRefs #194

### DIFF
--- a/src/lib/websocket.ts
+++ b/src/lib/websocket.ts
@@ -731,6 +731,7 @@ export function useWebSocket() {
 
       ws.onclose = (event) => {
         log.info(`Disconnected from Gateway: ${event.code} ${event.reason}`)
+        const wasHandshakeComplete = handshakeCompleteRef.current
         setConnection({ isConnected: false })
         handshakeCompleteRef.current = false
         stopHeartbeat()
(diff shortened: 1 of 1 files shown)


0xNyk:
Security/maintainer review: the one-file state-ordering fix is logically sound and I found no suspicious or expanded behavior. Local typecheck, 17 gateway URL tests, lint (0 errors), and the strict security scan pass on current main.\n\nBefore merge, please:\n1. update/rebase the branch onto current main (it is currently behind), and\n2. add a focused regression test that exercises an established connection closing after handshake and asserts reconnect stays on the last known-good URL instead of trying /gateway-ws or /gw.\n\nThe existing gateway-url tests only cover URL construction, so they do not prove this close-handler state transition. Once those are in, we can run the protected hosted gate and merge.

PR #913, opened 2026-07-29
feat: add Atlas Cloud direct provider

binyangzhu000-sudo:
## Summary

- add Atlas Cloud to the model catalog and integrations settings
- route Atlas models through the existing OpenAI-compatible direct dispatch path
- preserve nested Atlas model IDs and support `ATLASCLOUD_API_KEY`
- add focused catalog, routing, and request-shape coverage

## Validation

- `pnpm test` (178 files, 1547 tests)
- `pnpm typecheck`
- `pnpm lint`
- focused provider tests (22 tests)
- live Atlas model discovery: HTTP 200, 122 models, default model present
- live Atlas chat completions: HTTP 200
- `git diff --check`

`pnpm build` reached the Next.js production build but could not download the upstream Inter and JetBrains Mono files from `fonts.googleapis.com` because the connection reset. No Atlas-related compiler error was reported.

## Scope

No README, public documentation, logo, sponsor, credits, or partner content was changed.


0xNyk:
Audited clean — vetted against a `main` baseline with zero new dropper, secret, or invisible-unicode findings, and the PR text passed the injection scan.

Held rather than merged: this adds a model provider, which is an architecture and trust decision for the maintainer rather than an automated one, and branch protection requires `quality-gate` plus CodeQL — both currently held pending workflow approval for fork PRs.

binyangzhu000-sudo:
Rebuilt this branch on current `main` — it was conflicting after two upstream changes landed in the same code paths this PR touches.

**What conflicted, and how I resolved it**

1. `resolveDirectProvider()` → `pickProvider()`. I took the upstream name everywhere, including in the test.
2. A `minimax` direct provider was added at exactly the extension points this PR uses. Both providers are kept side by side — `DirectProvider` now lists both `'atlascloud'` and `'minimax'`, and each has its own catalog branch, prefix branch, strip-prefix entry, availability check and dispatch case. I verified this by enumerating all nine `DirectProvider` consumer sites and checking Atlas Cloud is present at each one, since a textually clean merge can still silently drop a case.

One deliberate reduction: the Atlas dispatch test previously called `callDirectly`, which upstream has since made non-exported. Rather than re-export an internal function for my test's benefit, I dropped that assertion — the routing test now matches the granularity of your own MiniMax test.

**Verification** — `vitest run src/lib/__tests__/`: **171 files, 1362 tests, all passing**. Baseline on unmodified `main` is 168/1336, so the delta is exactly this PR's additions and nothing regressed. Confirmed my case actually executes (`-t "Atlas Cloud"` → 1 passed).

On the merge decision itself — understood, and no push from me. You said adding a model provider is an architecture and trust call for a maintainer rather than an automated one, and that's your call to make. I've kept the branch mergeable so the option stays open; happy to close it if you'd rather not take on another provider.

Disclosure, in case it's not obvious from the diff: I work at Atlas Cloud.

PR #925, opened 2026-08-11
fix(security): harden env ignore, host allowlist, and logout cookies

dakota2499:
## Summary
- Ignore all `.env*` variants by default (including `.env.bak*`) so backup env files with secrets cannot be accidentally committed; keep `.env.example` and `.env.test` trackable.
- Make production host filtering fail-closed when `MC_ALLOWED_HOSTS` is empty (only implicit local hosts), unless `MC_ALLOW_ANY_HOST=1` is set.
- On logout, revoke and expire both `__Host-mc-session` and legacy `mc-session` cookies so HTTP→HTTPS transitions cannot leave a valid session behind.
- Add regression tests for the host gate and dual-cookie session helpers.

## Context
These changes came from a local security audit of Mission Control. They are small, high-value hardening fixes with unit test coverage.

## Test plan
- [x] `vitest run src/proxy.test.ts` (includes empty `MC_ALLOWED_HOSTS` fail-closed case)
- [x] `vitest run src/lib/__tests__/session-cookie.test.ts`
- [ ] CI green on this PR
- [ ] Manual: production-mode host with empty allowlist blocks unrelated Host header; localhost still works
- [ ] Manual: logout clears both cookie names

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,8 +12,12 @@ dist/
 # Misc
 .DS_Store
 *.pem
-.env*.local
-.env.production
+
+# Secrets — ignore all env variants by default (incl. .env.bak*, .env.local, etc.)
+# Re-allow only the intentionally committed sample/test files.
+.env*
+!.env.example
+!.env.test
 
 # npm (project uses pnpm)
 package-lock.json
@@ -29,7 +33,6 @@ yarn-error.log*
 # TypeScript
 *.tsbuildinfo
 next-env.d.ts
-.env
 .data/
 aegis/
 
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -1,3 +1,4 @@
+# pnpm 10+ reads these settings from this file (not package.json "pnpm" field).
 onlyBuiltDependencies:
   - '@parcel/watcher'
   - '@swc/core'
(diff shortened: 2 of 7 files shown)


0xNyk:
Reviewed in full. This is real security work and both fixes are correct.

**`.gitignore`** — the old `.env*.local` + `.env.production` + `.env` set left `.env.bak`, `.env.prod`, `.env.old` and similar untracked-but-not-ignored. Moving to `.env*` with an explicit `!.env.example` / `!.env.test` allowlist closes that, and allowlisting beats enumerating variants you have to keep guessing at.

**logout** — the old path destroyed a single parsed token, so a request presenting both the secure and legacy cookie names had one session survive logout. `parseAllMcSessionCookieTokens` + `expireAllMcSessionCookies` revokes both, and the added tests cover parse order and expiry options rather than just the happy path.

Audited: vetted against a `main` baseline — zero new dropper/secret/invisible-unicode findings introduced.

Not merging it myself: branch protection requires `quality-gate` and CodeQL, and those runs are held pending maintainer approval for fork PRs. Once approved and green this is ready. One cosmetic note — the branch is named `add-chatgpt-openai-integration`, which doesn't match the contents; worth a rename if you open follow-ups from it.

0xNyk:
Holding this PR (unsigned first-time-contributor commit on a leftover ChatGPT/OpenAI branch name). The intended harden is real — fail-closed production hosts, expire both session cookies on logout, ignore `.env*` except sample/test — and we are re-implementing it first-party on current `main` instead of merging this branch.

Thank you for the report and the tests. We will close this once the first-party PR lands.

0xNyk:
Superseded by first-party #939 on current main (fail-closed production hosts, expire both session cookies, `.env*` ignore). Thank you for the report and tests — we re-implemented rather than merging this FTC/ChatGPT-named branch.

PR #927, opened 2026-08-13
feat(i18n): add zh-tw (Traditional Chinese, Taiwan) locale

YTingWu:
# Summary
Adds a `zh-tw` locale (Traditional Chinese, Taiwan) alongside the existing Simplified Chinese (`zh`) locale. This is a Taiwan-specific translation using Taiwan tech-industry terminology, not a straight character conversion of `zh.json` — e.g. `帳戶` not `賬戶`, `代理` not `智能體`, `品質` not `質量`, `使用者`/`登入`/`設定`/`專案`/`伺服器`/`快取` throughout.

Also fixes `src/i18n/request.ts` Accept-Language matching to try an exact BCP-47 tag match (e.g. `zh-TW`) before falling back to the bare 2-letter prefix. Previously any `zh-*` header resolved to `zh` (Simplified) because the code truncated to 2 characters before matching; now a `zh-TW` browser correctly resolves to `zh-tw`.

# Risk Level
Low. Additive locale + a narrowly-scoped fallback-matching fix in `request.ts`; no changes to existing locale files, routing, or auth. Main failure mode would be a missed/incorrect translation string in `messages/zh-tw.json`, which is a content issue, not a functional one — the locale-parity test guards against structural drift (missing/extra/empty keys) going forward.

# Evidence
- `pnpm test` → full suite passes (1546/1546 unit tests including `src/i18n/locale-parity.test.ts`, 21/21 — up from 19, confirming `zh-tw.json` has exact key parity with `en.json` and no empty strings)
- `pnpm typecheck` → clean
- `pnpm lint` → clean
- `pnpm build` → succeeds
- `pnpm test:e2e` → 514/514 passing, including the updated `tests/i18n-language-switcher.spec.ts` (now asserts 11 languages and a new test that switches to `zh-tw` and confirms the login page renders `登入以繼續` / `使用者名稱` / `登入` in a real browser)

# Contribution Checklist
- [x] This PR has one reviewable purpose and no unrelated cleanup
- [x] Tests cover behavior changes and failure paths (locale-parity test + e2e switcher test extended for the new locale)
- [x] `pnpm lint`, `pnpm typecheck`, and relevant tests pass
- [x] Auth, data scope, command execution, and secret handling were reviewed when touched (not touched by this PR)
- [ ] Schema changes include forward migration and upgrade-path tests (n/a — no schema changes)
- [x] Public text, logs, fixtures, screenshots, and commits contain no secrets or private data

# Notes
Two sections (`gatewayControl`, `agentSquadPhase3`) and a handful of onboarding/skills strings are intentionally left untranslated in `zh-tw.json`, matching `zh.json` and every other existing locale (es/ja/ko/de/fr/pt/ru/ar) — this is pre-existing translation debt shared across all locales, not specific to this PR, so I kept `zh-tw` consistent with its siblings rather than scope-creeping into a cross-locale fix.


0xNyk:
Audited clean. I want to record this explicitly because the first scan looked alarming:

A full-tree vet of this PR head returned 56 CRITICAL findings, which would ordinarily be a REJECT. Running the identical scan against `main` returned **the same 56**, so every one is pre-existing and this PR introduces none. The findings are all false positives in any case — `AKIAIOSFODNN7EXAMPLE` in the repo's own secret-scanner tests, dynamic `import()` calls read as exec sinks, and zero-width characters inside `injection-guard.test.ts` that are deliberate test inputs.

So: no finding against this contribution. Blocked only on `quality-gate` and CodeQL, which are held pending maintainer approval for fork PRs.

